PR #3, opened 2025-07-02
refactor: silence a bunch of compiler warnings

akhi3030:
Either removing some code or marking the variables as unused.

Tested using: `cargo check --all-targets --tests --all-features`


src/consensus.rs
```diff
@@ -55,15 +55,11 @@ use votor::Votor;
 pub const SLOTS_PER_WINDOW: u64 = 4;
 /// Number of slots in each epoch.
 pub const SLOTS_PER_EPOCH: u64 = 18_000;
-/// Time bound assumed on network transmission delays during periods of synchrony.
-const DELTA: Duration = Duration::from_millis(400);
```
qkniep: Rather than deleting these constants, I would actually use them, e.g.:
`const DELTA_EARLY_TIMEOUT: Duration = DELTA.checked_mul(2).unwrap();`
and
`const DELTA_TIMEOUT: Duration = DELTA_EARLY_TIMEOUT.checked_add(DELTA_BLOCK).unwrap();`

To actually use `DELTA_TIMEOUT`, we could then also change `votor.rs` from having:
`tokio::time::sleep(DELTA_BLOCK)).await;`
to
`tokio::time::sleep(DELTA_TIMEOUT.saturating_sub(DELTA_EARLY_TIMEOUT)).await;`

qkniep (changes requested):

src/consensus.rs
akhi3030: Thanks, now updated.

PR #7, opened 2025-07-03
fix CI

qkniep:
At first the CI was missing the ping dataset, which has to be downloaded from AWS. Now the download and also caching it are part of the CI workflow. Adding that I broke the workflow. This PR is supposed to fix that.

### DIFF
--- a/src/disseminator/turbine.rs
+++ b/src/disseminator/turbine.rs
@@ -374,6 +374,7 @@ mod tests {
         tokio::time::sleep(Duration::from_millis(100)).await;
 
         // non-leaders should have received all shreds via Turbine
+        // TODO: flaky
         assert_eq!(*shreds_received.lock().await, 9 * TOTAL_SHREDS);
         task_leader.abort();
         for task in tasks {


qkniep:
Now I am kind of stumped. The tests now fail because parsing of the file `data/servers-2020-07-19.csv` (which is not the downloaded file, it's tracked as part of the repo!) fails with the error:
`thread 'main' panicked at src/network/simulated/ping_data.rs:38:41:
called `Result::unwrap()` on an `Err` value: Error(Deserialize { pos: Some(Position { byte: 97, line: 2, record: 1 }), err: DeserializeError { field: None, kind: Message("missing field `_name`") } })`

Locally this works for me. It should also not be line endings (we are only using macOS and Linux).

akhi3030:
Digging into the csv crate, this is an error being thrown by serde.  It seems to be issued from https://github.com/BurntSushi/rust-csv/blob/1.3.1/src/deserializer.rs#L683.  

One theory is that somehow it is looking for the field "_name" and not "name".  So it seems like when it parsed the first line, it somehow screwed that up.

I'll keep digging.

akhi3030:
Looks like the problem has to do with whitespace.  I could not reproduce it exactly but asking the csv reader to trim it seems to fix the problem.

PR #8, opened 2025-07-03
refactor: redefine `Shred` to remove some code duplication and accessor functions

akhi3030:
redefine `Shred` so that the common fields of data and coding shards are not duplicated.  This also allows us to remove some accessor functions.

### DIFF
--- a/src/disseminator/turbine.rs
+++ b/src/disseminator/turbine.rs
@@ -85,7 +85,7 @@ impl<N: Network> Turbine<N> {
     pub async fn send_shred_to_root(&self, shred: &Shred) -> Result<(), NetworkError> {
         // TODO: fix duplicate use indices between data and coding shreds
         let tree = self
-            .get_tree(shred.payload().slot, shred.index_in_slot())
+            .get_tree(shred.payload().slot, shred.payload().index_in_slot())
             .await;
         let root = tree.get_root();
         let msg = NetworkMessage::Shred(shred.clone());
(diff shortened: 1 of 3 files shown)


src/shredder.rs
```diff
@@ -100,141 +92,69 @@ impl Slice {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum ShredPayloadType {
```
qkniep, line 96: We also still have the types `DataShred` and `CodingShred`, which could be confusing. But I'm not immediately sure how to improve that.

src/disseminator/turbine.rs
```diff
@@ -84,7 +84,9 @@ impl<N: Network> Turbine<N> {
     /// Returns an error if the send operation on the underlying network fails.
     pub async fn send_shred_to_root(&self, shred: &Shred) -> Result<(), NetworkError> {
         // TODO: fix duplicate use indices between data and coding shreds
-        let tree = self.get_tree(shred.slot(), shred.index_in_slot()).await;
+        let tree = self
+            .get_tree(shred.payload().slot, shred.index_in_slot())
```
qkniep: somehow this feels weirdly asymmetric to have to access slot behind the `payload()` call but not `index_in_slot`

akhi3030: moving `index_in_slot` method into `ShredPayload` as discussed offline.

src/shredder.rs
akhi3030: also discussed offline.  Yes, this is indeed a bit confusing but it also enables useful functionality to return data and coding shards separately.  So it makes sense to keep it around for now.

qkniep (approved):
LGTM

PR #14, opened 2025-07-04
refactor: remove unused dependencies in Cargo.toml

akhi3030:
generated using https://github.com/est31/cargo-udeps.


qkniep:
We really have to fix those flaky tests, low specs of CI machines just make it unbearable. 😅

PR #17, opened 2025-07-04
Repair blocks only when necessary

### LINKED ISSUE #11: Repair blocks only when necessary
Right now, if we receive a `Notar` vote for a block hash and we don't have that block, we perform repair. This breaks the nice guarantee of Alpenglow that we only ever have to store at most a small constant number of blocks.

To fix this, we should only perform repair in the following two cases:
- We see a `NotarFallback` cert for the block hash.
- All conditions for safe-to-notar hold, except that we are missing the block and thus don't know if the parent has a `NotarFallback` cert.

qkniep:
This should fix #11

### DIFF
--- a/src/consensus/pool/slot_state.rs
+++ b/src/consensus/pool/slot_state.rs
@@ -90,6 +90,7 @@ pub struct SlotCertificates {
     pub(super) finalize: Option<FinalCert>,
 }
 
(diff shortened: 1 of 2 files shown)


akhi3030 (approved):
I don't completely grasp all the subtleties here yet but at a high level, it looks good.

PR #26, opened 2025-07-15
feat: receive txs from user when producing blocks

akhi3030:
When producing slices, we actually try to receive transactions from users to include into the slice.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -85,7 +85,7 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     disseminator: Arc<D>,
     /// Block repair protocol.
     repair: Arc<Repair<R>>,
-    /// network connection to receive transactions from users / rpc nodes, etc.
+    /// Network connection to receive transactions from clients.
     txs_receiver: T,
 
     /// Indicates whether the node is shutting down.
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -86,7 +86,7 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     /// Block repair protocol.
     repair: Arc<Repair<R>>,
     /// network connection to receive transactions from users / rpc nodes, etc.
-    txs_recver: T,
+    txs_receiver: T,
 
     /// Indicates whether the node is shutting down.
     cancel_token: CancellationToken,
@@ -110,7 +110,7 @@ where
         disseminator: D,
         repair_network: R,
         epoch_info: Arc<EpochInfo>,
-        txs_recver: T,
+        txs_receiver: T,
     ) -> Self {
         let cancel_token = CancellationToken::new();
         let (votor_tx, votor_rx) = mpsc::channel(1024);
@@ -163,7 +163,7 @@ where
             repair,
             cancel_token,
             votor_handle,
-            txs_recver,
+            txs_receiver,
         }
     }
 
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -21,6 +21,7 @@ mod blockstore;
 mod cert;
 mod epoch_info;
 mod pool;
+mod produce_block;
 mod vote;
 mod votor;
 
@@ -31,9 +32,7 @@ use std::{sync::Arc, time::Duration};
 use color_eyre::Result;
 use fastrace::Span;
 use fastrace::future::FutureExt;
-use log::{debug, info, trace, warn};
-use rand::rngs::SmallRng;
-use rand::{RngCore, SeedableRng};
+use log::{debug, trace, warn};
 use tokio::sync::{RwLock, mpsc};
 use tokio::time::sleep;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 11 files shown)


src/main.rs
```diff
@@ -133,19 +137,21 @@ fn create_test_nodes(count: u64) -> Vec<TestNode> {
         .iter()
         .map(|v| {
             let epoch_info = Arc::new(EpochInfo::new(v.id, validators.clone()));
-            let start_port = 3000 + v.id * 3;
-            let network = UdpNetwork::new(start_port as u16);
+            let start_port = 3000 + (v.id * 3) as u16;
```
qkniep: this needs to be `v.id * 4` now

tests/liveness.rs
```diff
@@ -103,13 +107,15 @@ fn create_test_nodes(count: u64) -> Vec<TestNode> {
             let all2all = TrivialAll2All::new(validators.clone(), networks.pop_front().unwrap());
             let disseminator = Rotor::new(networks.pop_front().unwrap(), epoch_info.clone());
             let repair_network = networks.pop_front().unwrap();
+            let txs_recver = networks.pop_front().unwrap();
```
qkniep: same, `txs_receiver`?

src/consensus/produce_block.rs
```diff
@@ -0,0 +1,145 @@
...
+            slice_index,
+            is_last,
+            merkle_root: None,
+            data,
+        },
+        cont_prod,
+    )
+}
+
+impl<A, D, R, T> Alpenglow<A, D, R, T>
```
qkniep, line 94: could be cleaner to have a struct for the leader / block production part, instead of continuing to directly implement this in `struct Alpenglow`? e.g. this struct would then own the `txs_receiver`, house the block production loop itself, and could later be extended with the scheduler... but we can also keep this change for a later PR

src/consensus.rs
```diff
@@ -86,18 +85,21 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network> {
     disseminator: Arc<D>,
     /// Block repair protocol.
     repair: Arc<Repair<R>>,
+    /// network connection to receive transactions from users / rpc nodes, etc.
+    txs_recver: T,
```
qkniep: nit: naming, maybe `txs_receiver` instead? a different name could also work, e.g. `client_socket`, `client_network`, `txs_socket`

akhi3030: Went with `txs_receiver`.  No strong preference.  Happy to change.

src/consensus/produce_block.rs
akhi3030: Agreed, I'll create an issue for this.

src/consensus/produce_block.rs
```diff
@@ -0,0 +1,205 @@
...
+enum Continue {
+    Continue { left: Duration },
+    Stop,
+}
+
+async fn produce_slice<T>(
+    txs_receiver: &T,
+    slot: Slot,
+    slice_index: usize,
+    parent: (Slot, Hash),
```
qkniep: passing `parent: (Slot, Hash)` here seems strange
maybe it should be `parent: Option<(Slot, Hash)>` because we usually only want to include the parent information in the first slice anyways AND we need to allow for the parent to change later for the optimistic handover (which could be implemented as having `Some(slot, hash)` for any slice except the first one)

src/consensus/produce_block.rs
```diff
@@ -0,0 +1,205 @@
...
+    sleep_duration: Duration,
```
qkniep: nit: maybe call this something like `time_left` or `time_left_for_block`?

src/consensus/produce_block.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use super::{Alpenglow, DELTA_BLOCK};
+
+enum Continue {
+    Continue { left: Duration },
+    Stop,
+}
+
+async fn produce_slice<T>(
+    txs_receiver: &T,
```
qkniep, line 23: I feel like `produce_slice` should not be the one directly reading from the network, e.g. replacing this with a channel might be a way but we can also keep that for alter.

src/consensus.rs
```diff
@@ -160,6 +163,7 @@ where
             repair,
             cancel_token,
             votor_handle,
+            txs_receiver,
```
qkniep, line 166: still not in love with the name `receiver` for a network interface / socket, but we can also keep it if we don't immediately find something better

src/consensus/produce_block.rs
```diff
@@ -20,23 +22,25 @@ enum Continue {
 async fn produce_slice<T>(
     txs_receiver: &T,
     slot: Slot,
-    slice_index: usize,
-    parent: (Slot, Hash),
+    slice_index: Either<(Slot, Hash), NonZeroUsize>,
```
qkniep: optimistic handover

src/consensus.rs
akhi3030: Yes, I cannot think of a better name right now either.  I somehow think that `receiver` makes a bit of sense as will only ever receive msgs here

src/consensus/produce_block.rs
akhi3030: Yeah, I think something like that makes sense and will help with testing too.  I think we can add something as a follow up.

src/consensus/produce_block.rs
akhi3030: Now fixed.

src/consensus/produce_block.rs
akhi3030: now handled a bit better.

PR #30, opened 2025-07-16
feat: extend blockstore to store more than one block per slot

### LINKED ISSUE #16: Prepare Blockstore for storing different blocks as part of repair
Right now, `Blockstore` is not yet entirely equipped for dealing with more than one block.

In addition to the `shreds` and `slices` fields, we probably want to have `repair_shreds` and `repair_slices`, both mapping from block hashes to one WIP set of shreds/slices.

qkniep:
This should fix #16.

### DIFF
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -5,19 +5,23 @@
 //!
 //!
 
+mod slot_block_data;
+
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
-use log::{debug, trace, warn};
+use either::Either;
+use log::debug;
 use tokio::sync::mpsc::Sender;
 
-use crate::crypto::signature::PublicKey;
-use crate::crypto::{Hash, MerkleTree};
-use crate::shredder::{self, RegularShredder, Shred, Shredder, Slice};
+use crate::crypto::Hash;
+use crate::shredder::{Shred, Slice};
 use crate::{Block, Slot};
 
 use super::epoch_info::EpochInfo;
 use super::votor::VotorEvent;
+
+use slot_block_data::SlotBlockData;
 
 /// Information about a block within a slot.
 #[derive(Clone, Copy, Debug)]
@@ -46,36 +50,6 @@ pub struct Blockstore {
     votor_channel: Sender<VotorEvent>,
     /// Information about all active validators.
     epoch_info: Arc<EpochInfo>,
-}
-
-///
-pub struct SlotBlockData {
-    ///
-    slot: Slot,
-    /// Stores raw shreds per slice, indexed by slot and slice index.
-    shreds: BTreeMap<usize, Vec<Shred>>,
-    /// Stores so far reconstructed slices, indexed by slot and slice index.
-    slices: BTreeMap<usize, Slice>,
-    ///
-    // repair_shreds: BTreeMap<(Hash, usize), Vec<Shred>>,
-    ///
-    // repair_slices: BTreeMap<(Hash, usize), Slice>,
-    /// Stores actual fully reconstructed block data for all relevant blocks.
-    blocks: BTreeMap<Hash, Block>,
-    /// Holds currently canonical block hash for a given slot number.
-    canonical: Option<Hash>,
-    /// Holds hashes of alternative blocks for a given slot number.
-    alternatives: Vec<Hash>,
-    /// Indicates for which slots we already received at least one shred.
-    first_shred_seen: bool,
-    /// Stores double-Merkle tree for each canonical block.
-    double_merkle_tree: Option<MerkleTree>,
-    /// Stores last slice index for each slot.
-    last_slice: Option<usize>,
-    /// Set of slots for which conflicting shreds have been seen (leader equivocated).
-    equivocated: bool,
-    /// Cache of previously verified Merkle roots.
-    merkle_root_cache: BTreeMap<usize, Hash>,
 }
 
 impl Blockstore {
@@ -110,31 +84,35 @@ impl Blockstore {
     ) -> Option<(Slot, BlockInfo)> {
         let slot = shred.payload().slot;
         let leader_pk = self.epoch_info.leader(slot).pubkey;
-        let (block_info, event) = self
+        let add_shred_result = self
             .slot_data_mut(slot)
             .add_shred(shred.clone(), check_equivocation, leader_pk)
             .await;
 
-        // potentially notify Votor of first slice
-        if let Some(event) = event {
-            self.votor_channel.send(event).await.unwrap();
+        match &add_shred_result? {
+            Either::Right(event) => {
+                // notify Votor of first slice
+                self.votor_channel.send(event.clone()).await.unwrap();
+                None
+            }
+            Either::Left(block_info) => {
+                // notify Votor of block and print block info
+                let event = VotorEvent::Block {
+                    slot,
+                    block_info: *block_info,
+                };
+                self.votor_channel.send(event).await.unwrap();
+                debug!(
+                    "reconstructed block {} in slot {} with parent {} in slot {}",
+                    &hex::encode(block_info.hash)[..8],
+                    slot,
+                    &hex::encode(block_info.parent_hash)[..8],
+                    block_info.parent_slot,
+                );
+
+                Some((slot, *block_info))
+            }
         }
-
-        // return early if no block was reconstructed
-        let block_info = block_info?;
-
-        // notify Votor of block and print block info
-        let event = VotorEvent::Block { slot, block_info };
-        self.votor_channel.send(event).await.unwrap();
-        debug!(
-            "reconstructed block {} in slot {} with parent {} in slot {}",
-            &hex::encode(block_info.hash)[..8],
-            slot,
-            &hex::encode(block_info.parent_hash)[..8],
-            block_info.parent_slot,
-        );
-
-        Some((slot, block_info))
     }
 
     /// Gives the canonical block hash for a given `slot`, if any.
@@ -224,167 +202,6 @@ impl Blockstore {
     }
 }
 
-impl SlotBlockData {
-    fn new(slot: Slot) -> Self {
-        Self {
-            slot,
-            shreds: BTreeMap::new(),
-            slices: BTreeMap::new(),
-            blocks: BTreeMap::new(),
-            canonical: None,
-            alternatives: Vec::new(),
-            first_shred_seen: false,
-            double_merkle_tree: None,
-            last_slice: None,
-            equivocated: false,
-            merkle_root_cache: BTreeMap::new(),
-        }
-    }
-
-    async fn add_shred(
-        &mut self,
-        shred: Shred,
-        check_equivocation: bool,
-        leader_pk: PublicKey,
-    ) -> (Option<BlockInfo>, Option<VotorEvent>) {
-        assert_eq!(shred.payload().slot, self.slot);
-        if check_equivocation && self.equivocated {
-            debug!("recevied shred from equivocating leader, not adding to blockstore");
-            return (None, None);
-        }
-        let slice = shred.payload().slice_index;
-        let index = shred.payload().index_in_slice;
-        let is_last_slice = shred.payload().is_last_slice;
-        let slice_shreds = self.shreds.entry(slice).or_default();
-
-        // check Merkle root and signature
-        let cached_merkle_root = self.merkle_root_cache.get(&slice);
-        if !shred.verify(&leader_pk, cached_merkle_root) {
-            debug!("dropping invalid shred");
-            return (None, None);
-        } else if cached_merkle_root.is_none() {
-            self.merkle_root_cache.insert(slice, shred.merkle_root);
-        } else if cached_merkle_root != Some(&shred.merkle_root) {
-            if !self.equivocated {
-                self.equivocated = true;
-                warn!("shreds show leader equivocation in slot {}", self.slot);
-            }
-            if check_equivocation {
-                return (None, None);
-            }
-        }
-
-        // store and handle this shred if and only if:
-        //   - it is not yet stored in the blockstore
-        //   - it is not (known to be) after the last slice
-        let exists = slice_shreds.iter().any(|s| {
-            s.payload().index_in_slice == index
-                && shred.payload_type.is_data() == s.payload_type.is_data()
-        });
-        let after_last = self.last_slice.is_some_and(|last_slice| slice > last_slice);
-        if exists || after_last {
-            return (None, None);
-        }
-        slice_shreds.push(shred);
-
-        // store last slice index, delete later shreds
-        if is_last_slice && self.last_slice.is_none() {
-            self.last_slice = Some(slice);
-
-            // delete shreds after last slice, if there are any
-            let keys_to_delete: Vec<_> = self
-                .shreds
-                .range(&slice + 1..)
-                .map(|(k, _)| k)
-                .copied()
-                .collect();
-            for key in keys_to_delete {
-                self.shreds.remove(&key);
-            }
-        }
-
-        // maybe send first shred notification
-        let event = match self.first_shred_seen {
-            true => None,
-            false => {
-                self.first_shred_seen = true;
-                Some(VotorEvent::FirstShred(self.slot))
-            }
-        };
-
-        // maybe reconstruct slice and block
-        if self.try_reconstruct_slice(slice) {
-            (self.try_reconstruct_block().await, event)
-        } else {
-            (None, event)
-        }
-    }
-
-    /// Reconstructs the slice if the blockstore contains enough shreds.
-    ///
-    /// Returns `true` if a slice was reconstructed, `false` otherwise.
-    fn try_reconstruct_slice(&mut self, slice: usize) -> bool {
-        let slice_shreds = self.shreds.get(&slice).unwrap();
-        if slice_shreds.len() < shredder::DATA_SHREDS || self.slices.contains_key(&slice) {
-            return false;
-        }
-
-        let reconstructed_slice = RegularShredder::deshred(slice_shreds).unwrap();
-        self.slices.insert(slice, reconstructed_slice);
-        trace!("reconstructed slice {} in slot {}", slice, self.slot);
-        true
-    }
-
-    /// Reconstructs the block if the blockstore contains all slices.
-    ///
-    /// Returns `Some(slot, block_info)` if a block was reconstructed, `None` otherwise.
-    /// In the `Some`-case, `block_info` is the [`BlockInfo`] of the reconstructed block.
-    async fn try_reconstruct_block(&mut self) -> Option<BlockInfo> {
-        if self.canonical.is_some() {
-            trace!("already have block for slot {}", self.slot);
-            return None;
-        }
-        let last_slice = self.last_slice?;
-        if self.slices.len() != last_slice + 1 {
-            trace!("don't have all slices for slot {} yet", self.slot);
-            return None;
-        }
-
-        // calculate double-Merkle tree & block hash
-        let merkle_roots = self
-            .slices
-            .values()
-            .map(|s| s.merkle_root.as_ref().unwrap())
-            .collect::<Vec<_>>();
-        let tree = MerkleTree::new(&merkle_roots);
-        let block_hash = tree.get_root();
-        self.double_merkle_tree = Some(tree);
-
-        // reconstruct block header
-        let first_slice = self.slices.get(&0).unwrap();
-        let parent_slot = u64::from_be_bytes(first_slice.data[0..8].try_into().unwrap());
-        let parent_hash = first_slice.data[8..40].try_into().unwrap();
-        self.canonical = Some(block_hash);
-        // TODO: reconstruct actual block content
-        let block = Block {
-            slot: self.slot,
-            block_hash,
-            parent: parent_slot,
-            parent_hash,
-            transactions: vec![],
-        };
-        let block_info = BlockInfo::from(&block);
-        self.blocks.insert(block_hash, block);
-
-        // clean up raw slices
-        for slice_index in 0..=last_slice {
-            self.slices.remove(&slice_index);
-        }
-
-        Some(block_info)
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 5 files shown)


qkniep:
before I continue with extending blockstore, did some refactoring

src/consensus/blockstore.rs
```diff
@@ -37,33 +38,44 @@ impl From<&Block> for BlockInfo {
...
+    epoch_info: Arc<EpochInfo>,
+}
+
+///
+pub struct SlotBlockData {
+    ///
+    slot: Slot,
     /// Stores raw shreds per slice, indexed by slot and slice index.
-    shreds: BTreeMap<(Slot, usize), Vec<Shred>>,
+    shreds: BTreeMap<usize, Vec<Shred>>,
```
akhi3030: Will there be at most 4 items in this vec?  If yes, in the short term, we can add that as a comment.  Longer term, it might be interesting to store the field as something like `[Option<Shred>; 4]` to save some indirection.  But that is definitely premature optimisation at this point.

src/consensus/blockstore.rs
```diff
@@ -37,33 +38,44 @@ impl From<&Block> for BlockInfo {
...
     /// Stores so far reconstructed slices, indexed by slot and slice index.
-    slices: BTreeMap<(Slot, usize), Slice>,
+    slices: BTreeMap<usize, Slice>,
```
akhi3030: Same comment as above.

I think we should also consider introducing (in a separate PR):
`struct SliceIndex(usize)`
`struct ShreadIndex(usize)`

This will improve readability and prevent accidentally mixing the two `usize`s.

src/consensus/blockstore.rs
```diff
@@ -37,33 +38,44 @@ impl From<&Block> for BlockInfo {
...
+    blocks: BTreeMap<Hash, Block>,
     /// Holds currently canonical block hash for a given slot number.
-    canonical: BTreeMap<Slot, Hash>,
+    canonical: Option<Hash>,
     /// Holds hashes of alternative blocks for a given slot number.
-    alternatives: BTreeMap<Slot, Vec<Hash>>,
+    alternatives: Vec<Hash>,
     /// Indicates for which slots we already received at least one shred.
-    first_shred_seen: BTreeSet<Slot>,
+    first_shred_seen: bool,
```
akhi3030: I wonder if this can be combined with the `shreds` field in the following way:
store `shreds: Option<BTreeMap<...>>,

If it is `None`, then we have not seen any shreds, if it is `Some`, then we have seen at least one.  Of course, the `BTreeMap` might be empty so we still have an edge case that is not handled.  But I think this reduces the number of invalid states somewhat.

src/consensus/blockstore.rs
```diff
@@ -166,168 +304,84 @@ impl Blockstore {
...
+                self.first_shred_seen = true;
+                Some(VotorEvent::FirstShred(self.slot))
+            }
+        };
 
         // maybe reconstruct slice and block
-        if self.try_reconstruct_slice(slot, slice) {
-            self.try_reconstruct_block(slot).await
+        if self.try_reconstruct_slice(slice) {
+            (self.try_reconstruct_block().await, event)
```
akhi3030: It seems like this function will never return `(Some(BlockInfo), None)`.  I would be tempted to introduce an enum for the return type. e.g.

```
enum AddShredResult {
  None,
  JustEvent(VotorEvent),
  BlockReconstructed(BlockInfo, VotorEvent),
}
```

src/consensus/blockstore.rs
```diff
@@ -166,168 +304,84 @@ impl Blockstore {
...
-        if self.stored_slices_for_slot(slot) != last_slice + 1 {
-            trace!("don't have all slices for slot {slot} yet");
+        let last_slice = self.last_slice?;
+        if self.slices.len() != last_slice + 1 {
+            trace!("don't have all slices for slot {} yet", self.slot);
             return None;
         }
 
         // calculate double-Merkle tree & block hash
-        let merkle_roots: Vec<_> = self
```
akhi3030, line 213: thank you 😇

src/consensus/blockstore.rs
```diff
@@ -105,31 +108,169 @@ impl Blockstore {
...
+
+    /// Generates a Merkle proof for the given `slice` within the given `slot`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the double-Merkle tree for the given `slot` does not exist.
+    #[must_use]
+    pub fn create_double_merkle_proof(&self, slot: Slot, slice: usize) -> Vec<Hash> {
+        let slot_data = self.slot_data(slot).unwrap();
+        let tree = slot_data.double_merkle_tree.as_ref().unwrap();
```
akhi3030: These `unwrap`s already existed so we don't necessarily need to fix them in this PR but they worry me a bit.  We should look into how we can eliminate them in future PRs.

src/consensus/blockstore.rs
```diff
@@ -37,33 +38,44 @@ impl From<&Block> for BlockInfo {
...
+    block_data: BTreeMap<Slot, SlotBlockData>,
+
+    /// Event channel for sending notifications to Votor.
+    votor_channel: Sender<VotorEvent>,
+    /// Information about all active validators.
+    epoch_info: Arc<EpochInfo>,
+}
+
+///
+pub struct SlotBlockData {
```
akhi3030: It will be great to move this into its own module separate from `BlockStore`.  They both have similar sounding function names and it makes it a bit harder to navigate the code.

src/consensus/blockstore.rs
qkniep: no, these `Vec`s contain up to `TOTAL_SHREDS_PER_SLICE` (currently 64) shreds each

src/consensus/blockstore.rs
qkniep: also using `usize` is generally considered bad form? at least this one style guide you shared said this

src/consensus/blockstore.rs
qkniep: true, or just checking `!shreds.is_empty()`, will remove this

src/consensus/blockstore.rs
akhi3030: > also using `usize` is generally considered bad form? at least this one style guide you shared said this

Yeah, if we expect to be running on different architectures in particular 32 and 64 bits, then it can be problematic.  I imagine that we will always run on 64 bits so I am not so worried about this issue.

We can add 

```
 assert_eq!(size_of::<u64>(), size_of::<usize>());
```

somewhere on top of `main()` and then freely use `usize`.

src/consensus/blockstore.rs
akhi3030: I meant that the map can have up to `TOTAL_SHREDS_PER_SLICE` entries but is there an upper limit on how many `Shred`s can be in the `Vec` in each entry?  I expected the answer to be 4 based on our discussion this morning.  But maybe something is still unclear to me.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,201 @@
...
+use crate::shredder::{self, RegularShredder, Shred, Shredder, Slice};
+use crate::{Block, Slot};
+
+use super::BlockInfo;
+
+///
+pub struct SlotBlockData {
+    ///
+    slot: Slot,
+    /// Stores raw shreds per slice, indexed by slot and slice index.
```
```suggestion
    /// Stores raw shreds per slice, indexed by slice index.
```

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub(super) shreds: BTreeMap<usize, Vec<Shred>>,
+    /// Stores so far reconstructed slices, indexed by slot and slice index.
```
```suggestion
    /// Stores so far reconstructed slices, indexed by slice index.
```

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,201 @@
...
+        self.slices.insert(slice, reconstructed_slice);
+        trace!("reconstructed slice {} in slot {}", slice, self.slot);
+        true
+    }
+
+    /// Reconstructs the block if the blockstore contains all slices.
+    ///
+    /// Returns `Some(slot, block_info)` if a block was reconstructed, `None` otherwise.
+    /// In the `Some`-case, `block_info` is the [`BlockInfo`] of the reconstructed block.
+    async fn try_reconstruct_block(&mut self) -> Option<BlockInfo> {
```
akhi3030: i don't think this function needs to be async.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,201 @@
...
+            canonical: None,
+            alternatives: Vec::new(),
+            double_merkle_tree: None,
+            last_slice: None,
+            equivocated: false,
+            merkle_root_cache: BTreeMap::new(),
+        }
+    }
+
+    pub async fn add_shred(
```
akhi3030: if try_reconstruct_block is not async, then I think this can also be not async.

src/consensus/blockstore.rs
```diff
@@ -105,230 +81,121 @@ impl Blockstore {
...
-        } else if cached_merkle_root != Some(&shred.merkle_root) {
-            if self.equivocated_slots.insert(slot) {
-                warn!("shreds show leader equivocation in slot {slot}");
+        let leader_pk = self.epoch_info.leader(slot).pubkey;
+        let event = self
+            .slot_data_mut(slot)
+            .add_shred(shred.clone(), check_equivocation, leader_pk)
+            .await;
+
+        match event.clone()? {
```
akhi3030:
```suggestion
        let event = event?;
        match &event {
```

I think then we can remove the `unwrap`s below maybe?  I haven't tested the theory though.

src/consensus/blockstore.rs
```diff
@@ -105,230 +81,121 @@ impl Blockstore {
...
-            let event = VotorEvent::FirstShred(slot);
-            self.votor_channel.send(event).await.unwrap();
-        }
-
-        // maybe reconstruct slice and block
-        if self.try_reconstruct_slice(slot, slice) {
-            self.try_reconstruct_block(slot).await
-        } else {
-            None
+            _ => unreachable!(),
```
akhi3030:
```suggestion
            ev => panic!("unexpected event {ev:?}"),
```

Will help with debugging in future.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,201 @@
...
+    ///
+    slot: Slot,
+    /// Stores raw shreds per slice, indexed by slot and slice index.
+    pub(super) shreds: BTreeMap<usize, Vec<Shred>>,
+    /// Stores so far reconstructed slices, indexed by slot and slice index.
+    pub(super) slices: BTreeMap<usize, Slice>,
+    ///
+    // repair_shreds: BTreeMap<(Hash, usize), Vec<Shred>>,
+    ///
+    // repair_slices: BTreeMap<(Hash, usize), Slice>,
```
```suggestion
```

src/consensus/blockstore/slot_block_data.rs
qkniep: good catch

src/consensus/blockstore.rs
qkniep: does not work, because `send()` takes the event by value and you cannot move `event` while it's still borrowed

akhi3030: hmm... so the following patch seems to `cargo check` for me.

```
diff --git a/src/consensus/blockstore.rs b/src/consensus/blockstore.rs
index ca1f45b..7b6fb22 100644
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -87,15 +87,16 @@ impl Blockstore {
             self.slot_data_mut(slot)
                 .add_shred(shred.clone(), check_equivocation, leader_pk);

-        match event.clone()? {
+        let event = event?;
+        match event {
             VotorEvent::FirstShred(_) => {
                 // notify Votor of first slice
-                self.votor_channel.send(event.unwrap()).await.unwrap();
+                self.votor_channel.send(event).await.unwrap();
                 None
             }
             VotorEvent::Block { slot, block_info } => {
                 // notify Votor of block and print block info
-                self.votor_channel.send(event.unwrap()).await.unwrap();
+                self.votor_channel.send(event).await.unwrap();
                 debug!(
                     "reconstructed block {} in slot {} with parent {} in slot {}",
                     &hex::encode(block_info.hash)[..8],
```

Not sure I understand why but I won't argue with the compiler.

src/consensus/blockstore/slot_block_data.rs
akhi3030: Looks like there are clippy lints that would catch such issues.  `cargo clippy` finds a lot of issues in the repo.  We should probably fix them then turn it on on CI.  I'll create an issue.

qkniep:
main batch of work for this PR is done now, clippy still complaining about some of the newly introduced errors not being handled properly (will do that later), and probably we want to move `BlockData` struct into its own file as well, but the main changes are ready to be looked at

src/consensus/blockstore.rs
```diff
@@ -100,163 +76,60 @@ impl Blockstore {
...
-
-        // maybe reconstruct slice and block
-        if self.try_reconstruct_slice(slot, slice) {
-            self.try_reconstruct_block(slot).await
-        } else {
-            None
+        let leader_pk = self.epoch_info.leader(slot).pubkey;
+        match self
+            .slot_data_mut(slot)
+            .add_shred_from_disseminator(shred.clone(), leader_pk)?
```
akhi3030: unnecessary clone?

src/consensus/blockstore.rs
```diff
@@ -100,163 +76,60 @@ impl Blockstore {
...
+    pub async fn add_shred_from_repair(
+        &mut self,
+        hash: Hash,
+        shred: Shred,
+    ) -> Result<Option<(Slot, BlockInfo)>, AddShredError> {
+        let slot = shred.payload().slot;
+        let leader_pk = self.epoch_info.leader(slot).pubkey;
+        match self
+            .slot_data_mut(slot)
+            .add_shred_from_repair(hash, shred.clone(), leader_pk)?
```
akhi3030: unnecessary clone?

src/consensus/blockstore.rs
```diff
@@ -100,163 +76,60 @@ impl Blockstore {
...
-        let block_info = BlockInfo::from(&block);
-        self.blocks.insert((slot, block_hash), block);
-
-        // clean up raw slices
-        for slice_index in 0..=*last_slice {
-            self.slices.remove(&(slot, slice_index));
+    async fn send_votor_event(&self, event: VotorEvent) -> Option<(Slot, BlockInfo)> {
+        match event {
+            VotorEvent::FirstShred(_) => {
+                // notify Votor of first slice
```
akhi3030: I don't think this comment adds a lot of additional context.

src/consensus/blockstore.rs
```diff
@@ -100,163 +76,60 @@ impl Blockstore {
...
+                self.votor_channel.send(event).await.unwrap();
+                None
+            }
+            VotorEvent::Block { slot, block_info } => {
+                // notify Votor of block and print block info
```
akhi3030: I don't think this comment adds a lot of additional context.

src/consensus/blockstore.rs
```diff
@@ -100,163 +76,60 @@ impl Blockstore {
...
+                    "reconstructed block {} in slot {} with parent {} in slot {}",
+                    &hex::encode(block_info.hash)[..8],
+                    slot,
+                    &hex::encode(block_info.parent_hash)[..8],
+                    block_info.parent_slot,
+                );
+
+                Some((slot, block_info))
+            }
+            ev => panic!("unexpected event {ev:?}"),
```
akhi3030, line 129: To me this signals that in future we should consider splitting up `VotorEvent` so that we do not have to handle unexpected events here.

src/consensus/blockstore.rs
```diff
@@ -265,70 +138,105 @@ impl Blockstore {
...
-        self.alternatives.get(&slot).map(|v| v.as_ref())
+    /// Gives the number of stored slices for a given `slot`.
+    pub fn stored_slices_for_slot(&self, slot: Slot) -> usize {
+        self.slot_data(slot).map_or(0, |s| s.canonical.slices.len())
+    }
+
+    /// Gives the number of stored shreds for a given `slot` (across all slices).
+    pub fn stored_shreds_for_slot(&self, slot: Slot) -> usize {
+        self.slot_data(slot).map_or(0, |s| {
+            s.canonical.shreds.values().map(Vec::len).sum::<usize>()
```
akhi3030:
```suggestion
            s.canonical.shreds.values().map(Vec::len).sum()
```

I think the compiler should be able to pick this up.

src/consensus/blockstore.rs
```diff
@@ -265,70 +138,105 @@ impl Blockstore {
...
+            .map_or(0, |s| s.canonical.shreds.get(&slice).map_or(0, Vec::len))
+    }
+
+    // /// Gives any relevant alternative block hashes for a given slot, if any.
+    // #[must_use]
+    // pub fn alternative_block_hashes(&self, slot: Slot) -> &[Hash] {
+    //     self.slot_data(slot)
+    //         .map_or(&[], |s| s.alternatives.as_ref())
+    // }
+
```
akhi3030:
```suggestion
```

Dead code?

src/consensus/blockstore.rs
```diff
@@ -265,70 +138,105 @@ impl Blockstore {
...
         tree.create_proof(slice)
     }
+
+    /// Deletes everything before the given `slot` from the blockstore.
+    pub fn prune(&mut self, slot: Slot) {
+        self.block_data = self.block_data.split_off(&slot);
+    }
+
+    ///
+    fn slot_data(&self, slot: Slot) -> Option<&SlotBlockData> {
```
akhi3030, line 218: I'd be tempted to remove this function as it is not doing a lot and the indirection adds some cognitive load.

src/consensus/blockstore.rs
```diff
@@ -265,70 +138,105 @@ impl Blockstore {
     /// Returns `None` if blockstore does not hold any version of the block yet.
     #[must_use]
     pub fn canonical_block_hash(&self, slot: Slot) -> Option<Hash> {
-        self.canonical.get(&slot).copied()
+        self.slot_data(slot)?
+            .canonical
+            .completed
+            .clone()
+            .map(|c| c.0)
```
akhi3030:
```suggestion
        self.slot_data(slot)?
            .canonical
            .completed
            .map(|c| c.0.clone())
```

so we do not end up cloning the `Block` unnecessarily.

src/consensus/blockstore.rs
```diff
@@ -265,70 +138,105 @@ impl Blockstore {
...
+                return Some(block);
+            }
+        } else {
+            return slot_data
+                .alternatives
+                .get(&hash)
+                .and_then(|data| data.completed.as_ref().map(|(_, block)| Some(block)))
+                .flatten();
+        }
+        return None;
```
akhi3030:
```suggestion
    match slot_data.canonical.completed.as_ref() {
        Some((h, block)) => {
            if *h == hash {
                return Some(block);
            }
            None
        }
        
        None => {
            slot_data
                .alternatives
                .get(&hash)
                .and_then(|data| data.completed.as_ref().map(|(_, block)| Some(block)))
                .flatten();
        }
    }
```

I somehow find this version a bit easier to read.  It is easier to understand which conditions hold when and what we are returning.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,272 @@
...
+        assert_eq!(shred.payload().slot, self.slot);
+        if self.equivocated {
+            debug!("recevied shred from equivocating leader, not adding to blockstore");
+            return Err(AddShredError::Equivocation);
+        }
+        let add_shred_result = self.canonical.check_shred_to_add(&shred, true, leader_pk);
+        if matches!(add_shred_result, Err(AddShredError::Equivocation)) {
+            self.equivocated = true;
+        }
+        add_shred_result?;
```
akhi3030, line 92: If `check_shred_to_add` ever starts returning something other than `()` in the `Ok` case, the compiler will help us catch this place to update.  Maybe we can write the above code as something like this:

```
match check_shred_to_add(...) {
  Ok(()) => Ok(self.canonical.add_valid_shred(shred)),
  Err(err) => {
    if let AddShredError::Equivocation = &err {
       self.equivocated = true;
    }
    Err(err)
  }
}
```

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,272 @@
...
+            slot,
+            completed: None,
+            shreds: BTreeMap::new(),
+            slices: BTreeMap::new(),
+            last_slice: None,
+            double_merkle_tree: None,
+            merkle_root_cache: BTreeMap::new(),
+        }
+    }
+}
```
akhi3030: combine with the other impl below?

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,272 @@
...
+
+impl BlockData {
+    ///
+    pub fn add_valid_shred(&mut self, shred: Shred) -> Option<VotorEvent> {
+        let slice_index = shred.payload().slice_index;
+        let is_last_slice = shred.payload().is_last_slice;
+        let is_first_shred = self.shreds.is_empty();
+        self.shreds.entry(slice_index).or_default().push(shred);
+
+        // store last slice index, delete later shreds
```
akhi3030: not sure I understand what "delete later shreds" means.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -0,0 +1,272 @@
...
+                })
+        } else {
+            None
+        }
+    }
+
+    fn check_shred_to_add(
+        &mut self,
+        shred: &Shred,
+        check_equivocation: bool,
```
akhi3030, line 177: For my knowledge, is checking for equivocation expensive or can it produce a false positive when repairing?  I am curious why we don't want to simply always check for it.

akhi3030 (commented):
generally looks good.  I think I'll do another pass next week after you'd had a chance to look at the feedback.

src/consensus/blockstore/slot_block_data.rs
qkniep: good catch

src/consensus/blockstore.rs
qkniep: good catch! thanks

src/consensus/blockstore/slot_block_data.rs
qkniep: improved comment and readability of the code + found a mistake as a result, which I fixed

src/consensus/blockstore.rs
qkniep: created an issue for this

src/consensus/blockstore.rs
qkniep: this was actually a bug, the logic should be like this: 
```
        if let Some((h, block)) = slot_data.canonical.completed.as_ref() {
            if *h == hash {
                return Some(block);
            }
        }
        slot_data
            .alternatives
            .get(&hash)
            .and_then(|data| data.completed.as_ref().map(|(_, block)| Some(block)))
            .flatten()
```

PR #32, opened 2025-07-17
refactor: separate errors for `add_cert` and `add_vote`

akhi3030:
I was expecting this change to be more impactful than it turned out to be.  I still think this is a good change to have long term but the status quo is perfectly fine for now.

We define different errors for the two functions so that they can evolve independently without impacting the callers of each others.  

With this change, we do not need to track what error types can be returned in comments as the expectation should be that all variants should be returnable.

### DIFF
--- a/src/consensus/pool.rs
+++ b/src/consensus/pool.rs
@@ -26,17 +26,28 @@ use super::{Cert, EpochInfo, SLOTS_PER_EPOCH, SLOTS_PER_WINDOW, Vote};
 use parent_ready_tracker::ParentReadyTracker;
 use slot_state::SlotState;
 
-/// Errors the Pool may throw when adding a vote or certificate.
+/// Errors the Pool may return when adding a vote.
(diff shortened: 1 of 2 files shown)


src/consensus/pool.rs
```diff
@@ -26,19 +26,30 @@ use super::{Cert, EpochInfo, SLOTS_PER_EPOCH, SLOTS_PER_WINDOW, Vote};
...
-    #[error("duplicate vote/cert")]
+    #[error("duplicate vote")]
     Duplicate,
     #[error("vote constitutes a slashable offence")]
     Slashable(SlashableOffence),
 }
 
+/// Errors the Pool may return when adding a certificate.
+#[derive(Clone, Copy, Debug, Error, PartialEq, Eq)]
+pub enum AddCertError {
```
qkniep, line 44: funnily enough, this reminds me that we could (and maybe should) add checking for slashable offences when adding certificates as well

akhi3030: haha... oh well, with this change we can take our time with it.

Issue #34, opened 2025-07-17
Run `cargo clippy` and enable it on CI


akhi3030:
Issue is self explanatory.  This will help address various issues.

qkniep:
Tried this before, there is already a `lint.sh` script for that, the vast amount of warnings regarding type casts between integer types is insane.

akhi3030:
Ah, thanks.  I'll take a look to see if we can fix them with some automation.

PR #35, opened 2025-07-18
chore: a bunch of clippy related fixes and enable it on CI

akhi3030:
Part of https://github.com/qkniep/alpenglow/issues/34.

This only enables an initial set of lints.  We probably want to expand which other set we would like to have.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -48,3 +48,6 @@ jobs:
 
       - name: Run tests
         run: ./test.sh ci
+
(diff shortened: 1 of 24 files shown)


akhi3030:
I want to make a few more changes to marking as draft.

akhi3030:
all relevant issues are fixed.

qkniep (approved):
looks good I think, at some point we should write the doc strings where you now deleted the placeholders (and probably in other places to)

PR #36, opened 2025-07-18
refactor: organise `shredder.rs` a bit

akhi3030:
- moves reed-solomon stuff into its own module
- introduces different error types to organise the errors a bit more

### DIFF
--- a/src/shredder.rs
+++ b/src/shredder.rs
@@ -16,17 +16,23 @@
 //! - [`Slice`] corresponds to a fixed number of shreds we perform erasure coding on.
 // TODO: split this file up
 
+mod reed_solomon;
+
 use aes::Aes128;
 use aes::cipher::{Array, KeyIvInit, StreamCipher};
 use ctr::Ctr64LE;
 use rand::{RngCore, rng};
-use reed_solomon_simd as rs;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


src/shredder.rs
```diff
@@ -16,18 +16,24 @@
 //! - [`Slice`] corresponds to a fixed number of shreds we perform erasure coding on.
 // TODO: split this file up
```
qkniep: remove this comment if we decide to not split this up further

src/shredder.rs
```diff
@@ -55,6 +61,39 @@ pub enum ShredderError {
...
+    fn from(err: ReedSolomonShredError) -> Self {
+        match err {
+            ReedSolomonShredError::TooMuchData => Self::TooMuchData,
+        }
+    }
+}
+
+/// Errors that may occur during shredding.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Error)]
+pub enum ShredError {
```
qkniep: nit: I would find the order `ShredError` before `DeshredError` more intuitive

akhi3030:
> at some point I was wondering if this file should be split into individual files for each different shredder implementation, did you actively decide against doing that?

I guess I implicitly decided against it as the different shredder implementations were not getting in my so much.  I think we can keep the code as is for now.  And revisit later when the file sees a lot of changes.

PR #38, opened 2025-07-19
Refactored `Transaction` out of `NetworkMessage`

MeerKatDev:
- Transaction is now its own type
 - I tried to follow conventional naming, and least amount of changes possible for this
 - not sure also whether I put those new elements in the right files, but the imports seem to be cleaner
 - implicit typing doesn't seem to be an option in this case


MeerKatDev:
Hi @qkniep should I close this? Are there any other issues I might help with?

Issue #4, opened 2025-07-03
Implement optimistic handover


qkniep:
The necessary changes for this are:
- [x] if block production loop sees last block of previous leader before `ParentReady`, start optimistically producing
- [ ] block production needs to be informed of new `ParentReady` events, potentially change parent of current block
- [x] measure block time for production only after `ParentReady`
- [ ] blockstore needs to be aware of this new block format, handle parents correctly

Some other notes of things to handle:
- [x] Introduce max slices per block.  This will also help simulate CUs.
- [x] we should only start delta block timer after parent ready event is seen.

akhi3030:
Cool.  any thoughts on how to be able to `await` on getting parent ready notifications?  

A straightforward idea would be to add a sender to `ParentReadyTracker` whose receiver is passed to `produce_blocks`.  But we want to be able to do it on a per slot level.  i.e. `ParentReadyTracker` might send parent_ready notification for slots that `produce_block` is not interested in.

So I am thinking is that `produce_block` should be able to "atomically" check if the slot has seen a parent ready already and if not then "await block" on a conditional variable on it.  Then `ParentReadyTracker` can notify the variable when it produces the first parent ready for that slot.

qkniep:
Makes sense.

Yes, maybe some atomic function on the `ParentReadyTracker`, like `check_parent_ready_or_subscribe` and it returns either the already existing ready parent or the receiver end of a one-shot channel. We can then have something like a `BTreeMap<Slot, Sender>` in the `ParentReadyTracker` and send the first `ParentReady` event for each slot to the corresponding sender.

akhi3030:
Thoughts on how to improve https://github.com/qkniep/alpenglow/pull/48.

Introduce two new structs: SliceHeader an SlicePayload.
- header contains things that go into each shred e.g. slot, slice_index, etc.
- payload contains the actual payload and that we do not want duplicated.

So SlicePayload can be defined roughly as:

```
struct SlicePayload {
  parent: Option<(ParentSlot, ParentHash)>,
  data: Vec<u8>,
}
```

And then the shredders only need to deal with these two structs and do not need to be updated when we make changes there.

PR #40, opened 2025-07-21
feat: make waiting for `ParentReady` `await`able

akhi3030:
Part of #4 

Adds the ability to be able to `await` for `ParentReady`.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -30,6 +30,7 @@ use std::time::Instant;
 use std::{sync::Arc, time::Duration};
 
 use color_eyre::Result;
+use either::Either;
 use fastrace::Span;
 use fastrace::future::FutureExt;
 use log::{debug, trace, warn};
@@ -41,7 +42,7 @@ use crate::crypto::{Hash, aggsig, signature};
 use crate::network::{Network, NetworkError, NetworkMessage};
 use crate::repair::{Repair, RepairMessage};
 use crate::shredder::Shred;
-use crate::{All2All, Disseminator, ValidatorInfo};
+use crate::{All2All, Disseminator, Slot, ValidatorInfo};
 
 pub use blockstore::{BlockInfo, Blockstore};
 pub use cert::Cert;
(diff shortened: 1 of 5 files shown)


src/consensus/pool/parent_ready_tracker/parent_ready_state.rs
```diff
@@ -0,0 +1,110 @@
...
+            IsReady::NotReady(_) => &[],
+            IsReady::Ready(block_ids) => block_ids,
+        }
+    }
+
+    pub(super) fn wait_for_parent_ready(
+        &mut self,
+    ) -> Either<&[BlockId], oneshot::Receiver<BlockId>> {
+        match self.is_ready {
+            IsReady::Ready(ref block_ids) => Either::Left(block_ids.as_slice()),
```
akhi3030: return just one.

src/consensus.rs
```diff
@@ -94,6 +95,74 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     votor_handle: tokio::task::JoinHandle<()>,
 }
 
+async fn wait_for_first_slot(
+    pool: Arc<RwLock<Pool>>,
+    blockstore: Arc<RwLock<Blockstore>>,
+    window: Slot,
+) -> Option<(Slot, Hash, bool)> {
```
qkniep, line 110: should probably have a comment explaining what the returned types refer to

src/consensus.rs
```diff
@@ -94,6 +95,74 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     votor_handle: tokio::task::JoinHandle<()>,
 }
 
+async fn wait_for_first_slot(
+    pool: Arc<RwLock<Pool>>,
+    blockstore: Arc<RwLock<Blockstore>>,
+    window: Slot,
```
qkniep, line 109: now that we want to be careful with types `window` should probably not be of type `Slot`, either it should be of a different type or we should just refer to it by its first slot

src/consensus.rs
```diff
@@ -94,6 +95,74 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
...
+) -> Option<(Slot, Hash, bool)> {
+    // Genesis
+    if window == 0 {
+        return Some((0, Hash::default(), false));
+    }
+
+    let first_slot_in_window = window * SLOTS_PER_WINDOW;
```
qkniep, line 116: todo

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -135,10 +126,21 @@ impl ParentReadyTracker {
...
     /// Returns list of all valid parents for the given slot, as of now.
+    /// The list can be empty.
     pub fn parents_ready(&self, slot: Slot) -> &[BlockId] {
         self.0
             .get(&slot)
-            .map_or(&[], |state| state.parents_ready.as_slice())
+            .map_or(&[], |state| state.ready_block_ids())
+    }
+
+    // If the slot has a parent ready, then returns it right away or else
```
qkniep: nit: comment should be docstring instead

src/consensus/pool/parent_ready_tracker/parent_ready_state.rs
```diff
@@ -0,0 +1,110 @@
...
+use smallvec::{SmallVec, smallvec};
+use tokio::sync::oneshot;
+
+use crate::crypto::Hash;
+
+use super::BlockId;
+
+/// Tracks the status of whether or not individual slots have parent ready.
+enum IsReady {
+    /// Do not have a parent ready for this slot.  Might have someone waiting
```
qkniep: style: I would usually put the line break between sentences in such cases

src/consensus/pool/parent_ready_tracker/parent_ready_state.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Holds the relevant state for a single slot.
+pub(super) struct ParentReadyState {
+    // XXX: consider making this field private
+    pub(super) skip: bool,
+    // we can potentially have multiple notar fallbacks per slot,
+    // but we optimize the common case where there will only be one
+    // XXX: consider making this field private
+    pub(super) notar_fallbacks: SmallVec<[Hash; 1]>,
+    // Do not make this field more visible.  Updating it must sometimes
```
qkniep: style: I have so far reserved proper punctuation and capitalization for docstrings, if we want to change that we can but I think we should be consistent

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -15,47 +15,37 @@
...
+use either::Either;
 use smallvec::SmallVec;
+use tokio::sync::oneshot;
 
 use crate::{Slot, consensus::SLOTS_PER_WINDOW, crypto::Hash};
 
-type BlockId = (u64, Hash);
+use parent_ready_state::ParentReadyState;
+
+pub type BlockId = (u64, Hash);
```
qkniep: use `Slot` here

src/consensus.rs
akhi3030: Agreed, I am actually going to fix in the other PR I'm working on where I introduce the window type.

src/consensus.rs
akhi3030: As discussed, will be addressed in the other PR.

PR #41, opened 2025-07-21
refactor: Introduce a Slot type

akhi3030:
Before, we stored the slot as a simple `u64` which is dangerous as it can be mixed up with other unrelated types.  This PR defines `Slot` as a `struct(u64)` to provide us with some type safety.

In general, I am not happy with the use of `SLOTS_PER_WINDOW` in so many tests.  But I am not sure if it makes sense to update them at this point in time.  We can discuss if we want to refactor them in this PR or a subsequent PR.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -51,8 +51,6 @@ pub use pool::{AddVoteError, Pool};
 pub use vote::Vote;
 use votor::Votor;
 
-/// Number of slots in each epoch.
-pub const SLOTS_PER_EPOCH: u64 = 18_000;
 /// Time bound assumed on network transmission delays during periods of synchrony.
 const DELTA: Duration = Duration::from_millis(400);
 /// Time the leader has for producing and sending the block.
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -135,7 +135,7 @@ async fn wait_for_first_slot(
             let handle = tokio::spawn(async move {
                 // PERF: These are burning a CPU. Can we use async here?
                 loop {
-                    let last_slot_in_prev_window = first_slot_in_window.last_slot_in_prev_window();
+                    let last_slot_in_prev_window = first_slot_in_window.prev();
                     if let Some(hash) = blockstore
                         .read()
                         .await
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -106,15 +106,13 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
 async fn wait_for_first_slot(
     pool: Arc<RwLock<Pool>>,
     blockstore: Arc<RwLock<Blockstore>>,
-    window: Slot,
+    first_slot_in_window: Slot,
 ) -> Option<(Slot, Hash, bool)> {
-    // genesis
-    if window == 0 {
-        return Some((0, Hash::default(), false));
+    if first_slot_in_window.is_genesis_window() {
+        return Some((Slot::new(0), Hash::default(), false));
     }
 
-    let first_slot_in_window = window * SLOTS_PER_WINDOW;
-    let last_slot_in_window = (window + 1) * SLOTS_PER_WINDOW - 1;
+    let last_slot_in_window = first_slot_in_window.last_slot_in_window();
 
     // if already have parent ready, return it, otherwise get a channel to await on
     let rx = {
@@ -141,18 +139,18 @@ async fn wait_for_first_slot(
             let handle = tokio::spawn(async move {
                 // PERF: These are burning a CPU. Can we use async here?
                 loop {
+                    let last_slot_in_prev_window = first_slot_in_window.last_slot_in_prev_window();
                     if let Some(hash) = blockstore
                         .read()
                         .await
-                        .canonical_block_hash(first_slot_in_window - 1)
+                        .canonical_block_hash(last_slot_in_prev_window)
                     {
-                        let slot = first_slot_in_window - 1;
                         debug!(
                             "optimistically building block on parent {} in slot {}",
                             &hex::encode(hash)[..8],
-                            slot
+                            last_slot_in_prev_window,
                         );
-                        return Some((slot, hash, false));
+                        return Some((last_slot_in_prev_window, hash, false));
                     }
                     if pool.read().await.finalized_slot() >= last_slot_in_window {
                         warn!(
@@ -313,7 +311,7 @@ where
     /// Keeps track of when consensus progresses, i.e., [`Pool`] finalizes new blocks.
     /// Triggers standstill recovery if no progress was detected for a long time.
     async fn standstill_loop(self: &Arc<Self>) -> Result<()> {
-        let mut finalized_slot = 0;
+        let mut finalized_slot = Slot::new(0);
         let mut last_progress = Instant::now();
         loop {
             let slot = self.pool.read().await.finalized_slot();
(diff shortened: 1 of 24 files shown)


src/consensus.rs
```diff
@@ -333,13 +331,13 @@ where
     /// Once all previous blocks have been notarized or skipped and the next
     /// slot belongs to our leader window, we will produce a block.
     async fn block_production_loop(&self) -> Result<()> {
-        for window in 0.. {
+        let mut first_slot_in_window = Slot::new(0);
+        loop {
```
akhi3030: I do not like this at all.  I tried implementing a `Slot::windows() -> Iterator<Slot>`.  However that conflicts with async code and produces a rabbit hole of errors.  Any thoughts on how to improve this are welcome.

src/consensus.rs
```diff
@@ -354,17 +352,21 @@ where
...
+            {
+                Some(res) => res,
+                None => continue,
+            };
 
             // produce blocks for all slots in window
             let mut block = parent;
             let mut block_hash = parent_hash;
-            for slot in first_slot_in_window..=last_slot_in_window {
+            for slot in first_slot_in_window.slots_in_window() {
```
akhi3030, line 364: Same issue as https://github.com/qkniep/alpenglow/pull/41/files#r2225812850 here.

src/consensus.rs
```diff
@@ -141,18 +139,18 @@ async fn wait_for_first_slot(
             let handle = tokio::spawn(async move {
                 // PERF: These are burning a CPU. Can we use async here?
                 loop {
+                    let last_slot_in_prev_window = first_slot_in_window.last_slot_in_prev_window();
```
qkniep: is this the only place where we use `last_slot_in_prev_window()`? because in this case it should simply be equivalent to `first_slot_in_window - 1`

qkniep: also, something like `first_slot_in_window.prev_window().last()` or `first_slot_in_window.window().prev().last()` using a double-ended iterator might be slightly more idiomatic, but those are details...

src/consensus/epoch_info.rs
```diff
@@ -32,9 +30,7 @@ impl EpochInfo {
     // TODO: actual stake-based pseudorandom leader schedule
     #[must_use]
     pub fn leader(&self, slot: Slot) -> &ValidatorInfo {
-        let window = slot / SLOTS_PER_WINDOW;
-        let leader_id = window % (self.validators.len() as u64);
-        self.validator(leader_id)
+        slot.leader(&self.validators)
```
qkniep: not sure it makes sense to move this implementation out of `EpochInfo` since this just a simple placeholder for now

src/consensus/votor.rs
```diff
@@ -245,16 +239,19 @@ impl<A: All2All + Sync + Send + 'static> Votor<A> {
     ///
     /// Panics if `slot` is not the first slot of a window.
     fn set_timeouts(&self, slot: Slot) {
-        assert_eq!(slot % SLOTS_PER_WINDOW, 0);
+        assert!(slot.is_start_of_window());
+        // TODO: set timeouts only once?
+        let window = slot.slots_in_window().into_iter().collect::<Vec<_>>();
```
qkniep: no need to collect here

src/slot.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn take(&self) -> u64 {
+        self.0
+    }
+
+    pub fn to_be_bytes(&self) -> [u8; 8] {
+        self.0.to_be_bytes()
+    }
+
+    pub fn slots_in_window(&self) -> Vec<Slot> {
```
qkniep: in most places it looks like we use `into_iter()` on this, maybe return `Iterator` here instead?

src/consensus.rs
qkniep: maybe we need to discuss this, I don't see the issue yet

src/slot.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Slot number type.
+#[derive(Clone, Copy, Debug, PartialOrd, Ord, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct Slot(u64);
+
+impl Slot {
+    pub fn new(slot: u64) -> Self {
+        Self(slot)
+    }
+
+    pub fn take(&self) -> u64 {
```
qkniep: maybe rename to something like `inner()` or `raw()`, for `take()` I have to think of the mutating [Option.take()](https://doc.rust-lang.org/std/option/enum.Option.html#method.take), which could be confusing

src/consensus.rs
akhi3030: I'll be in the office tomorrow, we can discuss in person.  The general issue is that, ideally we could write this as:

```
for first_slot_in_window in Slot::windows() {
  ...
}

impl Slot {
  fn windows() -> impl Iterator<Item=Slot> {
    // returns an infinite iterator that yields the first slot in each window.
  }
}
```

However, because we are using async here, the iterator can be sent across threads and therefore has to be `Send`.  And implementing `Send` on iterators gets pretty complicated pretty quickly.  Instead, I am using `loop` which is not as clean as the above approach.

qkniep: still not sure I understand, we can discuss tomorrow, but this compiles and runs just fine:
```rust
#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
struct Slot(u64);

impl Slot {
    pub fn new(slot: u64) -> Self {
        Slot(slot)
    }

    pub fn windows() -> impl Iterator<Item = Slot> + Send + Sync {
        (0..).map(|s| Slot::new(s * 4))
    }
}

#[tokio::main]
async fn main() {
    for slot in Slot::windows() {
        println!("{}", slot.0);
    }
}
```

src/consensus/votor.rs
akhi3030: Changed `slots_in_window()` to return an iterator.  Still need to collect here as otherwise we do not hold on to the function input `slot` for long enough.  Might still be a way to fix this though.

src/consensus.rs
akhi3030: this is not an issue anymore

src/consensus.rs
akhi3030: weird, I cannot reproduce the problem now.  And the simple iterator seems to work.  Must have been a problem between the keyboard and the monitor.

src/consensus/votor.rs
qkniep: I think the solution to this is just to take `self` by value in `slots_in_window()`, which is fine because `Slot` implements `Copy` anyways

akhi3030: Ah, good idea... that works.

src/consensus/epoch_info.rs
akhi3030: I was trying to hide the definition of SLOTS_PER_WINDOW but I ended up exposing it so I undid this change.

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -83,20 +83,25 @@ impl ParentReadyTracker {
...
         // find possible parents for future windows
         let mut potential_parents = SmallVec::<[BlockId; 1]>::new();
-        let start_of_window = slot.saturating_sub(slot % SLOTS_PER_WINDOW);
-        for s in (start_of_window..=slot).rev() {
+
+        let mut slots = slot
+            .slots_in_window()
+            .filter(|s| *s <= slot)
+            .collect::<Vec<_>>();
+        slots.reverse();
```
qkniep: `slots_in_window` could return a `DoubleEndedIterator` instead of just an `Iterator`, then we should be able to directly call `rev()` on the iterator

qkniep (changes requested):
looks good, I made one suggestion to make `slots_in_window()` even nicer to use

src/consensus.rs
akhi3030: should be fixed now.

src/consensus/pool/parent_ready_tracker.rs
akhi3030: good idea, now fixed.

qkniep (approved):
lgtm

PR #48, opened 2025-07-23
wip: store optional parent info in the slice

akhi3030:
This is to support optimistic handover.

### DIFF
--- a/src/shredder/reed_solomon.rs
+++ b/src/shredder/reed_solomon.rs
@@ -5,7 +5,7 @@
 
 use reed_solomon_simd as rs;
 
-use crate::Slot;
+use crate::{BlockId, Slot};
 
 use super::{
     CodingShred, DATA_SHREDS, DataShred, MAX_DATA_PER_SLICE, Shred, ShredPayload, ShredPayloadType,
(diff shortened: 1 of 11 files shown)


src/shredder/reed_solomon.rs
```diff
@@ -30,20 +30,30 @@ pub(super) fn reed_solomon_shred(
...
+    )
 }
 
 /// Splits the given data into `num_data` data shreds, then generates
 /// `num_coding` additional Reed-Solomon coding shreds.
-/// The slice-specific fields `slot`, `slice_index`, and `is_last` are included
-/// in every `DataShred` or `CodingShred`.
+/// The slice-specific fields `parent`, `slot`, `slice_index`, and `is_last`
+/// are included in every `DataShred` or `CodingShred`.
 pub(super) fn reed_solomon_shred_raw(
```
akhi3030, line 55: @qkniep: if I introduce the field in the naive way like I did here, then we will be sending the parent info in each shred in the slice.  So that will be 40 bytes extra per shred I suppose.  Is that problematic you think and something we should optimise?

qkniep: apart from the overhead, I'm not a big fan of passing this additional info all the way down to the Reed-Solomon calls, maybe already encoding it into the `data` at a higher level would be nicer?

akhi3030: makes sense.  Let me see how I can play around with this.

Issue #52, opened 2025-07-24
Add test for certificate too far in future


akhi3030:
We spotted and fixed a bug in https://github.com/qkniep/alpenglow/pull/50.  We should also add a test case that would've detected this bug.

PR #53, opened 2025-07-24
tests: add tests for `add_vote` and `add_cert` failure cases in Pool

qkniep:
This addresses #52

### DIFF
--- a/src/consensus/pool.rs
+++ b/src/consensus/pool.rs
@@ -473,8 +473,8 @@ mod tests {
     #[tokio::test]
     async fn handle_invalid_votes() {
         let (_, epoch_info) = generate_validators(11);
-        let (votor_tx, votor_rx) = mpsc::channel(1024);
-        let (repair_tx, repair_rx) = mpsc::channel(1024);
+        let (votor_tx, _votor_rx) = mpsc::channel(1024);
+        let (repair_tx, _repair_rx) = mpsc::channel(1024);
         let mut pool = Pool::new(epoch_info, votor_tx, repair_tx);
 
         let wrong_sk = SecretKey::new(&mut rand::rng());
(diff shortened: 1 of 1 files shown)


src/consensus/pool.rs
```diff
@@ -498,7 +498,10 @@ mod tests {
         assert!(!pool.is_notarized(0));
         for v in 0..11 {
             let vote = Vote::new_notar(0, Hash::default(), &sks[v as usize], v);
-            assert_eq!(pool.add_vote(vote).await, Ok(()));
+            match pool.add_vote(vote).await {
```
akhi3030: There are no changes to production code in this PR.  So is it the case that we expect one of two different things to happen here (`Ok(())` or `Err(AddVoteError::SlotOutOfBounds)`).  But so far all the tests only trigger the `Ok(())`?

src/consensus/pool.rs
```diff
@@ -917,4 +932,157 @@ mod tests {
...
+            pool.add_cert(Cert::Notar(notar_cert)).await,
+            Err(AddCertError::Duplicate)
+        );
+        assert_eq!(
+            pool.add_cert(Cert::Skip(skip_cert)).await,
+            Err(AddCertError::Duplicate)
+        );
+
+        drop(votor_rx);
+        drop(repair_rx);
```
akhi3030: for my knowledge really, why do we need to drop these?  Won't they get dropped when they go out of scope below?

akhi3030 (approved):

src/consensus/pool.rs
qkniep: yes, good catch, this was just a workaround to prevent "unused variable" errors, also assigning the receiver to `_` would immediately drop the receiver, now I settled on assigning to `_votor_rx` without the explicit `drop()` call

src/consensus/pool.rs
qkniep: reverted and moved to #55

PR #56, opened 2025-07-25
refactor: introduce `SliceHeader` and `SlicePayload` to hide info from the shredders

akhi3030:
Allows for deconstructing a `Slice` into `SliceHeader` and `SlicePayload`.  The header is included in each `Shred` while the payload is actually shredded into individual `Shred`s.

This change will allow us to make changes to `Slice` without having as much disruption in the shredders.

Part of #4.


qkniep (approved):
lgtm, only changed some doc strings

PR #60, opened 2025-07-30
feat: track parent separately in `Slice`

akhi3030:
Part of #4.  We now track parent in the slice separately from the actual data payload.  This will make it easier to indicate when a slice contains a parent or not.

### DIFF
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -263,7 +263,12 @@ mod tests {
     fn create_random_block(slot: Slot, num_slices: usize) -> Vec<Slice> {
         let mut slices = Vec::new();
         for slice_index in 0..num_slices {
-            let payload = create_random_slice_payload(MAX_DATA_PER_SLICE);
+            let parent = if slice_index == 0 {
+                Some((Slot::new(0), Hash::default()))
+            } else {
+                None
+            };
+            let payload = create_random_slice_payload(parent, MAX_DATA_PER_SLICE);
             let header = SliceHeader {
                 slot,
                 slice_index,
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -238,9 +238,9 @@ mod tests {
     use crate::shredder::{
         DATA_SHREDS, MAX_DATA_PER_SLICE, RegularShredder, Shredder, TOTAL_SHREDS,
     };
+    use crate::slice::{SliceHeader, create_random_slice_payload};
 
     use color_eyre::Result;
-    use rand::RngCore;
     use tokio::sync::mpsc;
 
     fn test_setup(tx: Sender<VotorEvent>) -> (SecretKey, Blockstore) {
@@ -262,17 +262,14 @@ mod tests {
 
     fn create_random_block(slot: Slot, num_slices: usize) -> Vec<Slice> {
         let mut slices = Vec::new();
-        let mut rng = rand::rng();
         for slice_index in 0..num_slices {
-            let mut buf = vec![0u8; MAX_DATA_PER_SLICE];
-            rng.fill_bytes(&mut buf);
-            slices.push(Slice {
+            let payload = create_random_slice_payload(MAX_DATA_PER_SLICE);
+            let header = SliceHeader {
                 slot,
                 slice_index,
                 is_last: slice_index == num_slices - 1,
-                merkle_root: None,
-                data: buf,
-            });
+            };
+            slices.push(Slice::from_parts(header, payload, None));
         }
         slices
     }
--- a/src/consensus/produce_block.rs
+++ b/src/consensus/produce_block.rs
@@ -14,7 +14,7 @@ use crate::MAX_TRANSACTION_SIZE;
 use crate::crypto::Hash;
 use crate::network::NetworkMessage;
 use crate::shredder::{MAX_DATA_PER_SLICE, RegularShredder, Shredder};
-use crate::slice::Slice;
+use crate::slice::{Slice, SliceHeader, SlicePayload};
 use crate::{All2All, Disseminator, Slot, network::Network};
 
 use super::{Alpenglow, DELTA_BLOCK};
@@ -28,27 +28,24 @@ async fn produce_slice<T>(
     txs_receiver: &T,
     slot: Slot,
     slice_index: Either<(Slot, Hash, usize), NonZeroUsize>,
-    time_left: Duration,
+    mut time_left: Duration,
 ) -> (Slice, Continue)
 where
     T: Network + Sync + Send + 'static,
 {
     const_assert!(MAX_DATA_PER_SLICE >= MAX_TRANSACTION_SIZE);
-    let (mut data, slice_index) = match slice_index {
+    let (parent, slice_index) = match slice_index {
         Either::Left((parent_slot, parent_hash, slice_index)) => {
-            let mut data = Vec::with_capacity(MAX_DATA_PER_SLICE);
-            // pack parent information in first slice
-            data.extend_from_slice(&parent_slot.inner().to_be_bytes());
-            data.extend_from_slice(&parent_hash);
-            let slice_capacity_left = MAX_DATA_PER_SLICE.checked_sub(data.len()).unwrap();
-            assert!(slice_capacity_left >= MAX_TRANSACTION_SIZE);
-            // FIXME: add support for optimistic handover. parent can change in middle of block production.
-            assert_eq!(slice_index, 0);
-            (data, slice_index)
+            (Some((parent_slot, parent_hash)), slice_index)
         }
-        Either::Right(ind) => (Vec::with_capacity(MAX_DATA_PER_SLICE), ind.get()),
+        Either::Right(ind) => (None, ind.get()),
     };
-    let mut left = time_left;
+
+    let parent_encoded_len = bincode::serde::encode_to_vec(parent, bincode::config::standard())
+        .unwrap()
+        .len();
+    let mut slice_capacity_left = MAX_DATA_PER_SLICE - parent_encoded_len;
+    let mut txs = Vec::new();
 
     let cont_prod = loop {
         let start_time = Instant::now();
@@ -62,12 +59,12 @@ where
                     Err(err) => panic!("Unexpected error {err}"),
                     Ok(msg) => match msg {
                         NetworkMessage::Transaction(tx) => {
-                            let mut bytes = bincode::serde::encode_to_vec(&tx, bincode::config::standard())
+                            let tx = bincode::serde::encode_to_vec(&tx, bincode::config::standard())
                                 .expect("serialization should not panic");
-                            data.append(&mut bytes);
-                            let slice_capacity_left = MAX_DATA_PER_SLICE.checked_sub(data.len()).unwrap();
+                            slice_capacity_left = slice_capacity_left.checked_sub(tx.len()).unwrap();
+                            txs.push(tx);
                             if slice_capacity_left < MAX_TRANSACTION_SIZE {
-                                break Continue::Continue { left };
+                                break Continue::Continue { left: time_left };
                             }
                         }
                         msg => {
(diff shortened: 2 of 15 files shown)


src/consensus/produce_block.rs
```diff
@@ -84,16 +79,13 @@ where
...
-            data,
-        },
-        cont_prod,
-    )
+    let header = SliceHeader {
+        slot,
+        slice_index,
+        is_last,
+    };
+    let payload = SlicePayload::from(data);
```
akhi3030: We are decoding here only to encode later and then decode again.  This could be improved but not sure if necessarily in this PR.

src/consensus/produce_block.rs
```diff
@@ -34,20 +34,15 @@ where
...
-            assert_eq!(slice_index, 0);
-            (data, slice_index)
+            (Some((parent_slot, parent_hash)), slice_index)
         }
-        Either::Right(ind) => (Vec::with_capacity(MAX_DATA_PER_SLICE), ind.get()),
+        Either::Right(ind) => (None, ind.get()),
     };
+    let mut data = Vec::with_capacity(MAX_DATA_PER_SLICE);
+    let mut parent = bincode::serde::encode_to_vec(parent, bincode::config::standard()).unwrap();
+    data.append(&mut parent);
```
qkniep: could instead directly `bincode::serde::encode_into_slice(&mut data)` here

src/slice.rs
```diff
@@ -83,24 +96,23 @@ pub(crate) struct SliceHeader {
...
-        Self { data }
-    }
-
-    /// Returns the size of the payload in bytes.
-    pub(crate) fn len(&self) -> usize {
-        let SlicePayload { data } = self;
-        data.len()
+impl From<SlicePayload> for Vec<u8> {
+    fn from(payload: SlicePayload) -> Self {
+        bincode::serde::encode_to_vec(payload, bincode::config::standard()).unwrap()
```
qkniep, line 117: bit unfortunate that we need to allocate an entirely new `Vec` here, would be great if we could reuse the memory from `data`, but probably not possible with `bincode`

src/slice.rs
```diff
@@ -83,24 +96,23 @@ pub(crate) struct SliceHeader {
...
     }
+}
 
-    /// Returns an iterator that iterates over the payload, `chunk_size` bytes at a time.
-    pub(crate) fn chunks(&self, chunk_size: usize) -> impl Iterator<Item = &[u8]> {
-        self.data.chunks(chunk_size)
+impl From<Vec<u8>> for SlicePayload {
+    fn from(payload: Vec<u8>) -> Self {
+        assert!(payload.len() <= MAX_DATA_PER_SLICE);
+        bincode::serde::decode_from_slice(&payload, bincode::config::standard())
```
qkniep: same here as above

src/consensus/produce_block.rs
qkniep: I don't have a better idea here either right now, I think we can leave it for now

akhi3030:
>     * `shredder::reed_solomon::tests::shred_too_much_data`: this is because we have assertions how that fire if you try to put more than MAX_DATA_PER_SLICE in a slice.  I propose we remove this test.

I suppose another way to handle this is to pass in a valid `SliceHeader` and a arbitrarily large payload.  I'll update the PR.

src/slice.rs
```diff
@@ -84,25 +98,75 @@ pub(crate) struct SliceHeader {
...
+fn num_bytes_set(mut val: usize) -> usize {
+    let mut cnt = 0;
+    while val != 0 {
+        val /= 256;
+        cnt += 1;
     }
+    cnt
+}
+
+// XXX: This is only used in test and benchmarking code.  Ensure it is only compiled when we are testing or benchmarking.
```
qkniep, line 135: `#[cfg(test)]`?

akhi3030:
Good point.  I think I managed to fix both the tests now.

src/slice.rs
```diff
@@ -24,36 +25,49 @@ pub struct Slice {
...
 }
 
 impl Slice {
-    /// Creates a [`Slice`] from raw payload bytes and the metadata extracted from a shred.
-    #[must_use]
-    pub(crate) const fn from_parts(data: Vec<u8>, any_shred: &Shred) -> Self {
+    pub(crate) fn from_parts(
+        header: SliceHeader,
+        payload: SlicePayload,
+        merkle_root: Option<[u8; 32]>,
```
qkniep, line 40: why is the Merkle root not part of the header?

src/slice.rs
```diff
@@ -24,36 +25,49 @@ pub struct Slice {
...
+    pub parent: Option<(Slot, Hash)>,
     /// Payload bytes.
     pub data: Vec<u8>,
 }
 
 impl Slice {
-    /// Creates a [`Slice`] from raw payload bytes and the metadata extracted from a shred.
-    #[must_use]
-    pub(crate) const fn from_parts(data: Vec<u8>, any_shred: &Shred) -> Self {
+    pub(crate) fn from_parts(
```
qkniep, line 37: docstring

src/slice.rs
```diff
@@ -24,36 +25,49 @@ pub struct Slice {
     pub is_last: bool,
     /// Merkle root hash over all shreds in this slice.
     pub merkle_root: Option<Hash>,
+    pub parent: Option<(Slot, Hash)>,
```
qkniep, line 30: docstring

src/slice.rs
```diff
@@ -84,25 +98,75 @@ pub(crate) struct SliceHeader {
...
+    let mut rng = rng();
+    rng.fill_bytes(&mut data);
+    bincode::serde::encode_into_slice(data, &mut payload[used..], bincode::config::standard())
+        .unwrap();
+
+    payload.into()
+}
+
+// XXX: This is only used in test and benchmarking code.  Ensure it is only compiled when we are testing or benchmarking.
+pub fn create_random_slice(desired_size: usize) -> Slice {
```
qkniep, line 170: docstring

src/consensus/blockstore.rs
```diff
@@ -262,17 +262,14 @@ mod tests {
...
+            let payload = create_random_slice_payload(MAX_DATA_PER_SLICE);
+            let header = SliceHeader {
                 slot,
                 slice_index,
                 is_last: slice_index == num_slices - 1,
-                merkle_root: None,
-                data: buf,
-            });
+            };
+            slices.push(Slice::from_parts(header, payload, None));
```
qkniep, line 277: It should not be a valid block if the first slice does not have a parent.

Related bug: See `try_reconstruct_block()` in `blockstore/slot_block_data.rs` lines 256ff, still tries to extract parent info from `data`.

src/slice.rs
```diff
@@ -84,25 +98,75 @@ pub(crate) struct SliceHeader {
...
+        println!("from: ret.data.len={} bytes={}", ret.data.len(), bytes);
+        // XXX: add an assertion that return bytes is equal to teh size of payload.
+        ret
     }
+}
 
-    /// Returns an iterator that iterates over the payload, `chunk_size` bytes at a time.
-    pub(crate) fn chunks(&self, chunk_size: usize) -> impl Iterator<Item = &[u8]> {
-        self.data.chunks(chunk_size)
+// Returns the number of bytes that are actually set in the `val`
```
qkniep: this comment should be a dosctring

qkniep: also, maybe it should be called "nonzero" and not "set"

src/shredder/reed_solomon.rs
```diff
@@ -128,41 +126,55 @@ mod tests {
...
+        let header = SliceHeader {
+            slot: Slot::new(0),
+            slice_index: 0,
+            is_last: true,
+        };
+        let payload = vec![0];
+        let (data, coding) =
+            reed_solomon_shred(header, payload.clone(), DATA_SHREDS, DATA_SHREDS).unwrap();
+        let shreds = take_and_map_enough_shreds(data, coding);
+        let restored = reed_solomon_deshred(&shreds, DATA_SHREDS, DATA_SHREDS).unwrap();
```
qkniep: we don't need the duplicate code here if we let `shred_deshred_restore()` take the `Vec<u8>` instead

src/network/simulated.rs
```diff
@@ -150,15 +148,13 @@ mod tests {
...
         let mut shreds = Vec::new();
         for i in 0..2 {
-            let mut data = vec![0; MAX_DATA_PER_SLICE];
-            rng.fill_bytes(&mut data);
-            let slice = Slice {
+            let payload = create_random_slice_payload(MAX_DATA_PER_SLICE);
+            let header = SliceHeader {
                 slot: Slot::new(0),
                 slice_index: i,
                 is_last: i == 4,
```
qkniep: oops
```suggestion
                is_last: i == 1,
```

src/consensus/produce_block.rs
```diff
@@ -62,38 +59,39 @@ where
...
                         }
                     },
                 }
-                left = left.saturating_sub(Instant::now() - start_time);
+                time_left = time_left.saturating_sub(Instant::now() - start_time);
             }
         }
     };
 
+    let txs = bincode::serde::encode_to_vec(&txs, bincode::config::standard())
```
qkniep, line 92: doesn't this add another `usize` to the encoding that we don't account for in `slice_capacity_left`?

src/consensus/produce_block.rs
```diff
@@ -62,38 +59,39 @@ where
...
+                                break Continue::Continue { left: time_left };
                             }
                         }
                         msg => {
                             panic!("Unexpected msg {msg:?}");
                         }
                     },
                 }
-                left = left.saturating_sub(Instant::now() - start_time);
+                time_left = time_left.saturating_sub(Instant::now() - start_time);
```
qkniep: nit: we don't account for all the elapsed time, since we set the `start_time` only at the beginning of the loop and we don't keep any buffer for the `bincode` encoding at the end

qkniep (commented):
great job, looks mostly good, but I left quite a few minor comments

src/consensus/blockstore.rs
akhi3030: Thanks, good spot.  I updates the code to check for the parent info in the first slice.  We should do more work here though to check that whenever we reconstruct a slice with index==1, then it has parent info otherwise it is dropped and similar check when we try to reconstruct a block.  I'll create an issue to handle this in a separate PR.

akhi3030: https://github.com/qkniep/alpenglow/issues/79

src/slice.rs
akhi3030: Looks like we had defined `SliceHeader` to not contain the merkle root in an earlier PR.  I think we did it like this so that we can use it inside the definition of `ShredPayload`.  It seems to make sense to me to not include merkle root in the header for this reason.  WDYT?

src/consensus/produce_block.rs
akhi3030: fixed.

src/consensus/produce_block.rs
akhi3030: I understood the second part of the comment but not sure I understood the first.  Does https://github.com/qkniep/alpenglow/pull/60/commits/957e36af20dc8fd5bfa5b8559474bb7b760afe43 address it?

akhi3030: I don't have a good solution for the second part of the concern here.  I propose we just add a todo for now and leave it as a future problem.

src/slice.rs
akhi3030: Sadly doesn't work... code gated by `#[cfg(test)]` is not visible to benchmarking code.  

AI suggested `#[cfg(any(test, feature = "bench"))]` which is not even a valid suggestion but essentially we need something like that.  An ability to enable a function (or the test_utils module) if testing or the bench feature are enabled.  But I cannot figure out the right incarnation and my vibe coding skills are failing me here.  

Any help is appreciated.

src/consensus/produce_block.rs
qkniep: that does capture some time that is not accounted for, I meant we don't capture what's happening before the loop, sure it's not a lot, that's why it's only a nitpick, but there is at least one call to bincode involving a memory allocation before the loop

qkniep: one way I could see to avoid such issues, is to not mutate `time_left` and instead only set `start_time` at the beginning of the function and then always calculate `time_left_now = time_left.saturating_sub(Instant::now() - start_time)`

src/slice.rs
qkniep: It seems it's kind of the opposite: the Merkle root only appears in `struct Shred` outside of `ShredPayload`. I think it makes sense, because it only can be added after shredding the slice and is thus different from all the other information in the slice (header and payload), which is part of what is initially being shredded. One way we could maybe improve that in the future is to have separate types for a slice before shredding and after de-shredding.

akhi3030: Yes that makes sense, then maybe we will not need an Option to store it which will allow us to have stronger type guarantee about when the merkle root should be present.  I'll create an issue to handle this in a separate PR.

akhi3030: #80

src/consensus/produce_block.rs
akhi3030: Ah now I understand.  Yes, your proposal is much cleaner.  Should be addressed in https://github.com/qkniep/alpenglow/pull/60/commits/97b97d9307fcbca4b48013592fd4900dc299e960.

Issue #62, opened 2025-08-01
`DELTA_TIMEOUT` seems redundant


akhi3030:
https://github.com/qkniep/alpenglow/blob/main/src/consensus.rs#L62.

`DELTA_TIMEOUT` is defined as `DELTA_EARLY_TIMEOUT + DELTA_BLOCK`.

https://github.com/qkniep/alpenglow/blob/main/src/consensus/votor.rs#L251.

The only place where it is used, we subtract `DELTA_EARLY_TIMEOUT` from it.

Based on my understanding, I think we want to use just `DELTA_EARLY_TIMEOUT` in votor.rs?

PR #63, opened 2025-08-01
fix: timeouts

qkniep:
This addresses #62 and fixes various minor problems with the timeouts:
- [x] inconsistent naming with the paper
- [x] https://github.com/qkniep/alpenglow/blob/main/src/consensus/votor.rs#L251
- [x] `2 * DELTA` too aggressive for Rotor
- [x] 400ms for `DELTA` potentially too high
- [ ] ~`3 * DELTA` conservative for Rotor~ (will be addressed separately, see #66)
- [ ] ~crashed leader timeout needs buffer for first slice~ (will be addressed separately, see #65)

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -52,7 +52,7 @@ pub use vote::Vote;
 use votor::Votor;
 
 /// Time bound assumed on network transmission delays during periods of synchrony.
-const DELTA: Duration = Duration::from_millis(200);
+const DELTA: Duration = Duration::from_millis(250);
 /// Time the leader has for producing and sending the block.
 const DELTA_BLOCK: Duration = Duration::from_millis(400);
 /// Timeout to use when we haven't seen any shred from the leader's block.
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -61,9 +61,7 @@ const DELTA: Duration = Duration::from_millis(400);
 const DELTA_BLOCK: Duration = Duration::from_millis(400);
 /// Timeout to use when we haven't seen any shred from the leader's block.
 /// This is used to skip honest but crashed leaders faster.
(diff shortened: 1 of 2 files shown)


src/consensus.rs
```diff
@@ -52,14 +52,11 @@ pub use vote::Vote;
...
+const DELTA: Duration = Duration::from_millis(250);
 /// Time the leader has for producing and sending the block.
 const DELTA_BLOCK: Duration = Duration::from_millis(400);
-/// Timeout to use when we haven't seen any shred from the leader's block.
-/// This is used to skip honest but crashed leaders faster.
-const DELTA_EARLY_TIMEOUT: Duration = DELTA.checked_mul(2).unwrap();
-/// Timeout to use when we have seen at least one shred from the leader's block.
-const DELTA_TIMEOUT: Duration = DELTA_EARLY_TIMEOUT.checked_add(DELTA_BLOCK).unwrap();
+/// Base timeout for when leader's first slice should arrive if they sent it immediately.
+const DELTA_TIMEOUT: Duration = DELTA.checked_mul(3).unwrap();
```
akhi3030, line 59: This is much better than what we had before.  

I would've expected to see a constant of something like "typical_time_to_create_first_slice" and `DELTA_TIMEOUT` defined as DELTA + this constant.

Another bit not being considered here is something about maximum clock drift between nodes.  Is that something that we should account for maybe?

akhi3030 (approved):
Approving as things are better than previously.  Left some comments on potential comments though.

src/consensus.rs
qkniep: The first part is basically #65, right?

And regarding clock drift, I really wouldn't care about it at the moment, drift for quartz clocks is on the order of ~0.002% whereas our constants like DELTA already have multiple percentage points of leeway.

Issue #51, opened 2025-07-23
Fix slow tests


qkniep:
Some tests in the category marked as "slow" tests seem to stall forever, specifically:
- `disseminator::rotor::sampling_strategy::tests::turbine_sampler`, and
- `disseminator::rotor::sampling_strategy::tests::turbine_sampler_real_world`

PR #68, opened 2025-08-01
fix: slow tests

qkniep:
This addresses #51 and fixes broken liveness tests.

### DIFF
--- a/src/disseminator/rotor/sampling_strategy.rs
+++ b/src/disseminator/rotor/sampling_strategy.rs
@@ -193,8 +193,8 @@ impl Clone for DecayingAcceptanceSampler {
 ///
 /// The goal is to distribute the required work for validators as in Turbine.
 /// Specifically, it should respect the same upper bound on the amount of work,
-/// that is, for `v` validators and given `turbine_fanout` any validator should
-/// be sampled no more than with probability `turbine_fanout / v`.
+/// that is, for `v` validators and given `fanout` any validator should
+/// be sampled no more than with probability `fanout / v`.
 #[derive(Clone)]
 pub struct TurbineSampler {
     fanout: usize,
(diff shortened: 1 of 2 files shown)


src/disseminator/rotor/sampling_strategy.rs
```diff
@@ -210,18 +211,16 @@ impl TurbineSampler {
...
     // TODO: support more than 2 levels of Turbine?
     #[must_use]
     pub fn new_with_fanout(mut validators: Vec<ValidatorInfo>, turbine_fanout: usize) -> Self {
         let total_stake: Stake = validators.iter().map(|v| v.stake).sum();
 
-        // calculate expected work for each validator
+        // calculate expected work for each validator (only excess over leader work)
         let mut expected_work = vec![0.0; validators.len()];
         for leader in &validators {
             let validators_left = validators.len() - 1;
```
akhi3030: nit: this value doesn't change in the each iteration of the loop, right.  Maybe consider defining it outside the loop.  That will help reduce the cognitive load on the reader.

There are some additional definitions further below like the other `validators_left`, `root_work`, `stake_left` which do not change in each iteration of the loop.  As a reader, I think it helps if they are defined outside of the loop then.

qkniep: ah yes, will move those around accordingly

src/disseminator/rotor/sampling_strategy.rs
```diff
@@ -189,14 +189,15 @@ impl Clone for DecayingAcceptanceSampler {
     }
 }
 
-/// A sampler that simulates the probability distribution in Turbine for Rotor.
+/// A sampler that simulates the probability distribution of Turbine for Rotor.
 ///
 /// The goal is to distribute the required work for validators as in Turbine.
 /// Specifically, it should respect the same upper bound on the amount of work,
 /// that is, for `v` validators and given `turbine_fanout` any validator should
 /// be sampled no more than with probability `turbine_fanout / v`.
```
akhi3030:
```suggestion
/// that is, for `v` validators and given `fanout` any validator should
/// be sampled no more than with probability `fanout / v`.
```

akhi3030 (approved):

PR #69, opened 2025-08-04
fix: Reed-Solomon padding

qkniep:
Reed-Solomon shredding and deshredding currently breaks for various slice data sizes (including empty slices and uneven number of bytes per shred). This was not caught by any tests.

Added some tests with variously sized slices to catch problems like these. Implemented padding to make those tests pass.

If I understand the requirements of the `reed-solomon-simd` crate correctly, the minimum padding we need to do is to pad payload data to a multiple of `2 * DATA_SLICES`. The alternative would be to simply always pad to the full `MAX_DATA_PER_SLICE`. We also have to account for storing the integer indicating the number of bytes of padding that was added.

### DIFF
--- a/src/shredder/reed_solomon.rs
+++ b/src/shredder/reed_solomon.rs
@@ -4,7 +4,6 @@
 //! Implements Reed-Solomon shreding and deshreding.
 
 use reed_solomon_simd as rs;
-use static_assertions::const_assert;
 use thiserror::Error;
 
 use super::{
@@ -42,8 +41,8 @@ pub(super) fn reed_solomon_shred(
 
     // add padding
     // TODO: use a padding scheme that can support larger slices
-    const_assert!(2 * DATA_SHREDS < 256);
-    let padding_bytes = (2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS)) as u8;
+    let padding_bytes = u8::try_from(2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS))
+        .expect("cannot fit number of padding bytes in u8");
     payload
         .data
         .resize(payload.len() + padding_bytes as usize, padding_bytes);


src/shredder/reed_solomon.rs
```diff
@@ -4,35 +4,52 @@
...
 ) -> Result<(Vec<DataShred>, Vec<CodingShred>), ReedSolomonShredError> {
     if payload.len() > MAX_DATA_PER_SLICE {
         return Err(ReedSolomonShredError::TooMuchData);
     }
 
-    let shred_bytes = payload.len() / DATA_SHREDS;
+    // add padding
+    // TODO: use a padding scheme that can support larger slices
+    const_assert!(2 * DATA_SHREDS < 256);
+    let padding_bytes = (2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS)) as u8;
```
akhi3030: nit: the assertion helps with reading the code but I think as a reader it will be even better to not use `as` here.  `as` will wrap around if the data does not fit in the `u8`.  Instead I would suggest using `try_from().unwrap()`.

src/shredder/reed_solomon.rs
```diff
@@ -4,35 +4,52 @@
...
+    mut payload: SlicePayload,
     num_data: usize,
     num_coding: usize,
 ) -> Result<(Vec<DataShred>, Vec<CodingShred>), ReedSolomonShredError> {
     if payload.len() > MAX_DATA_PER_SLICE {
         return Err(ReedSolomonShredError::TooMuchData);
     }
 
-    let shred_bytes = payload.len() / DATA_SHREDS;
+    // add padding
```
akhi3030, line 42: Doesn't need to be as part of this PR but maybe adding a small ascii art picture of what the padding would look like would help future readers a lot.

src/shredder/reed_solomon.rs
```diff
@@ -4,35 +4,52 @@
...
+    // TODO: use a padding scheme that can support larger slices
+    const_assert!(2 * DATA_SHREDS < 256);
+    let padding_bytes = (2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS)) as u8;
+    payload
+        .data
+        .resize(payload.len() + padding_bytes as usize, padding_bytes);
+    assert!(payload.len() <= MAX_DATA_PER_SLICE_AFTER_PADDING);
+
+    let shred_bytes = payload.len().div_ceil(DATA_SHREDS);
```
akhi3030, line 51: I am fine with this code.  But just to improve my understanding, I believe that payload.len() should be a multiple of DATA_SHREDS at this point in time, no?

src/shredder/reed_solomon.rs
```diff
@@ -86,17 +103,124 @@ pub(super) fn reed_solomon_deshred(
...
+
+    fn random_slice(num_bytes: usize) -> (SliceHeader, SlicePayload) {
+        let mut rng = rand::rng();
+        let header = SliceHeader {
+            slot: Slot::new(rng.random::<u64>()),
+            slice_index: rng.random::<u64>() as usize,
+            is_last: rng.random(),
+        };
+        let payload = SlicePayload {
+            // parent: rng.random(),
```
akhi3030:
```suggestion
```

src/slice.rs
```diff
@@ -70,7 +70,7 @@ impl Slice {
 /// Struct to hold all the header payload of a [`Slice`].
 ///
 /// This is included in each [`Shred`] after shredding.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
akhi3030: Do we need these here and below?  I like to introduce these only as needed because they just slow down compile times.

src/shredder/reed_solomon.rs
qkniep: yes, so going back to basic (floor) division here should not change anything here

PR #72, opened 2025-08-06
fix: main/alpenglow binary crashes immediately

qkniep:
Right now, simply running `./run.sh` crashes immediately.

This PR fixes all these and makes this simple local test run again.

### DIFF
--- a/src/slot.rs
+++ b/src/slot.rs
@@ -83,3 +83,20 @@ impl Display for Slot {
         write!(f, "{}", self.0)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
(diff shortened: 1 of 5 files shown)


qkniep:
ready now

src/slot.rs
```diff
@@ -28,7 +28,7 @@ impl Slot {
 
     /// Returns an infinite iterator that yields the first slot in each window.
     pub fn windows() -> impl Iterator<Item = Self> {
-        (0..).map(Self)
```
akhi3030, line 31: ouch, that is such a silly bug.  I really should've written some tests when I wrote this code 😢

Issue #71, opened 2025-08-05
Improve test coverage


qkniep:
Running `cargo llvm-cov --workspace --ignore-filename-regex='(benches|src/bin)' --release -- --include-ignored --test-threads=1` right now already gives a coverage of 84.8%. This is much better than what Codecov currently tells us based on CI, because CI does not run "ignored" tests.

Some files with bad coverage are:
- [x] `consensus/blockstore.rs` (#74)
- [x] `consensus/blockstore/slot_block_data.rs` (#74)
- [x] `consensus/votor.rs` (#113)
- [x] `crypto/aggsig.rs` (#73)
- [x] `crypto/hash.rs` (#73)
- [x] `crypto/signature.rs` (#73)
- [x] `disseminator/rotor/sampling_strategy.rs` (#75)
- [x] `logging.rs` (#105)
- [x] `network.rs` (#112)
- [x] `network/simulated/ping_data.rs` (#76)
- [x] `network/simulated/stake_distribution.rs` (#76)
- [ ] ~`network/tcp.rs`~
- [x] `repair.rs` (#106)

qkniep:
Test coverage when running the full `cargo llvm-cov --html --workspace --ignore-filename-regex='(benches|src/bin)' --release -- --include-ignored --test-threads=1` now surpasses 90%!
<img width="1033" height="902" alt="Screenshot 2025-08-20 at 14 09 57" src="https://github.com/user-attachments/assets/11196e1e-78b4-40dd-b998-ad7a0ab89c36" />

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/issues/74#issuecomment-3206059213_

PR #73, opened 2025-08-07
refactor: add more tests to `crypto` module

qkniep:
This is part of #71.

### DIFF
--- a/src/crypto/hash.rs
+++ b/src/crypto/hash.rs
@@ -45,3 +45,62 @@ pub fn hash_all(data: &[&[u8]]) -> Hash {
 pub fn truncate(hash: Hash) -> ShortHash {
     hash[..16].try_into().expect("wrong length")
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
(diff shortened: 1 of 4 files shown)


src/crypto/hash.rs
```diff
@@ -45,3 +45,62 @@ pub fn hash_all(data: &[&[u8]]) -> Hash {
...
+
+    const HASH_ITERATIONS: u64 = 100_000;
+
+    #[test]
+    fn deterministic() {
+        let hash1 = hash(&[0; 32]);
+        let hash2 = hash(&[0; 32]);
+        assert_eq!(hash1, hash2);
+
+        for i in 0..HASH_ITERATIONS {
```
akhi3030, line 63: does this test run as part of normal CI?  Would 100K iterations slow it down a lot?

akhi3030: Similar comments about tests below.

qkniep: this ran in <2s on CI, so it seems ok

PR #74, opened 2025-08-07
fix: shred indexing and more tests to `blockstore` module

qkniep:
This is part of #71.

While adding tests to the `blockstore` module, I also rediscovered and had to fix an old problem, that needed to be fixed for the new tests to pass: The `index_in_slice` for shreds was previously going e.g. 0..32 for data shreds and 0..32 for coding shreds again, instead coding shred indices should be offset (e.g. 0..32 for data shreds and 32..64 for coding shreds). This way this index uniquely refers to a shred, can be used directly in repair, everywhere in blockstore and for the Merkle trees. Only the underlying reed-solomon implementation still requires the old indexing, so when interfacing with it after shredding and before deshredding we have to translate the indexing.


qkniep:
I rediscovered and had to fix an old problem, that needed to be fixed for the new tests to pass: The `index_in_slice` for shreds was previously going e.g. 0..32 for data shreds and 0..32 for coding shreds again, instead coding shred indices should be offset (e.g. 0..32 for data shreds and 32..64 for coding shreds). This way this index uniquely refers to a shred, can be used directly in repair, everywhere in blockstore and for the Merkle trees. Only the underlying reed-solomon implementation still requires the old indexing, so when interfacing with it after shredding and before deshredding we have to translate the indexing.

src/consensus/blockstore.rs
```diff
@@ -160,6 +163,18 @@ impl Blockstore for BlockstoreImpl {
...
 
+    /// Stores a new shred from repair in the blockstore.
+    ///
+    /// This shred is stored in a spot associated with the given block`hash`.
+    /// For shreds obtained through block dissemination, `add_shred_from_disseminator`
+    /// should be used instead.
+    /// Compared to that function, this one does not check for leader equivocation.
+    ///
+    /// Reconstructs the corresponding slice and block if possible and necessary.
+    /// If the added shred belongs to the last slice, all later shreds are deleted.
```
akhi3030: do you mean shreds for later slices?

src/consensus/blockstore.rs
```diff
@@ -267,10 +283,47 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn store_slice() -> Result<()> {
+        let slot = Slot::genesis().next();
+        let (tx, _rx) = mpsc::channel(100);
+        let (sk, mut blockstore) = test_setup(tx);
+        assert!(blockstore.slot_data(slot).is_none());
+
+        // generate two slices for slot picked above
+        let slice = create_random_block(slot, 1)[0].clone();
```
akhi3030: the comment says 2 slices but the argument is for 1?

src/consensus/blockstore.rs
```diff
@@ -267,10 +283,47 @@ mod tests {
         (sk, BlockstoreImpl::new(Arc::new(epoch_info), tx))
     }
 
+    #[tokio::test]
+    async fn store_slice() -> Result<()> {
```
akhi3030: tangential: do we prefer tests to return a `Result` or tests that just assert the conditions they want and return nothing?  I see both flavours in the code base.  I think I have been adding mainly the second type so for.

akhi3030 (commented):
This PR makes me wonder if there is some cleanup we could do by introducing something like a `ShredIndex` type.  I haven't thought through the idea completely yet though.

LGTM.

src/consensus/blockstore.rs
qkniep: yes, updated the comment

src/consensus/blockstore.rs
qkniep: I just return `Result` here and in other places to be able to use the `?` operator instead of `unwrap()` for errors that are not part of the main test logic

PR #75, opened 2025-08-07
refactor: add more tests to `sampling_strategy` module

qkniep:
This is part of #71.

### DIFF
--- a/src/disseminator/rotor/sampling_strategy.rs
+++ b/src/disseminator/rotor/sampling_strategy.rs
@@ -755,6 +755,14 @@ mod tests {
         let sampled_set: HashSet<_> = sampled.iter().copied().collect();
         assert_eq!(sampled_set.len(), 100);
 
+        // heavy node sampled at most max_samples times
+        let mut validators = create_validator_info(100);
+        validators[0].stake = 10_000;
+        let sampler = DecayingAcceptanceSampler::new(validators, 5.0);
+        let sampled = sampler.sample_multiple(100, &mut rand::rng());
+        let sampled0 = sampled.into_iter().filter(|v| *v == 0).count();
+        assert!(sampled0 <= 5);
+
         // max_samples = inf equivalent to sampling with replacement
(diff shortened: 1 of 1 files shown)


src/disseminator/rotor/sampling_strategy.rs
```diff
@@ -231,7 +242,7 @@ impl TurbineSampler {
                 let root_work = (turbine_fanout as f64).min(validators_left as f64);
                 expected_work[root.id as usize] += prob * root_work;
                 let stake_left = stake_left - root.stake;
-                let validators_left = validators_left - turbine_fanout;
+                let validators_left = validators_left.saturating_sub(turbine_fanout);
```
akhi3030, line 245: This and all the validators_left calculations above are making assumptions about the minimum length of the validators input.  Should we add some assertions at the top of the function to help with readability?

src/disseminator/rotor/sampling_strategy.rs
```diff
@@ -469,6 +483,9 @@ impl FaitAccompli1Sampler<StakeWeightedSampler> {
 }
 
 impl<F: SamplingStrategy> SamplingStrategy for FaitAccompli1Sampler<F> {
+    // TODO: this is just a bad placeholder and should probably not be unimplemented
```
akhi3030, line 486: which bits are unimplemented?  I didn't get that.

src/disseminator/rotor/sampling_strategy.rs
qkniep: With the `saturating_sub` this should be correct now, for any number of validators.

src/disseminator/rotor/sampling_strategy.rs
qkniep: this should say "should probably not be implemented", because `sample()` for an individual validator makes no sense on these sampling strategies

akhi3030: got it, lgtm then.

Issue #31, opened 2025-07-17
Repair ancestors of finalized block


qkniep:
If we finalize a new block we need to initiate repair for all missing ancestors.

PR #78, opened 2025-08-08
wip: repair ancestors of finalized blocks

qkniep:
This addresses #31.

### DIFF
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -278,7 +278,7 @@ impl RepairRequest {
         }
     }
 
-    /// Returns the block hash this response refers to.
+    /// Returns the block hash this request refers to.
     #[must_use]
     pub const fn block_hash(&self) -> Hash {
         match self {


src/repair.rs
```diff
@@ -37,6 +42,7 @@ pub enum RepairRequest {
     SliceRoot(Slot, Hash, usize),
     /// Request for shred, identified by block hash, slice index and shred index.
     Shred(Slot, Hash, usize, usize),
+    // TODO: remove or replace with variant that includes proof
     Parent(Slot, Hash),
```
akhi3030, line 46: This is mainly for my understanding:

This is asking for the parent of the block identified by slot and hash, right?  Assuming that we are only ever repairing finalized blocks, do we need to include the hash?  Wouldn't it be sufficient to just send the slot?

I suppose it is good to send the hash so that if an honest receiver of the request has not finalized the block yet but it still has the parent of the block identified by hash, it can still reply to the request.  Otherwise, it will be forced to not respond?

src/repair.rs
```diff
@@ -181,28 +215,46 @@ impl<N: Network> Repair<N> {
...
                     .await
-                    .add_shred_from_repair(req.block_hash(), shred)
+                    .add_shred_from_repair(block_hash, shred)
                     .await
                     .unwrap();
             }
-            RepairResponse::Parent(_, parent_slot, parent_hash) => {
+            RepairResponse::Parent(req, parent_slot, parent_hash) => {
+                let RepairRequest::Parent(_, _) = req else {
+                    warn!("response to mismatching request type");
```
akhi3030: for debugging purpose, might be useful to output the req we got back.  Assuming the sender is not malicious and just buggy.

src/repair.rs
```diff
@@ -181,28 +215,46 @@ impl<N: Network> Repair<N> {
...
+                    return;
+                };
                 let block_info = BlockInfo {
                     hash: block_hash,
                     parent_slot,
                     parent_hash,
                 };
                 self.pool.write().await.add_block(slot, block_info).await;
+
+                // request repair of the parent block if necessary
```
akhi3030, line 243: This seems to be the core of the PR?  We requested a parent and after receiving it, we check if we have that block and if we don't we initiate another request for its parent and keep doing that till we we have repaired it all the way back to some block we have.

src/repair.rs
```diff
@@ -73,18 +84,36 @@ impl<N: Network> Repair<N> {
...
+    pub async fn repair_loop(&self) {
+        // only this loop uses the receiver
+        let mut guard = self.repair_channel.1.lock().await;
+        while let Some((slot, hash)) = guard.recv().await {
+            self.repair_block(slot, hash).await;
+        }
+    }
+
     /// Starts repair process for the block specified by `slot` and `block_hash`.
     pub async fn repair_block(&self, slot: Slot, block_hash: Hash) {
```
akhi3030, line 118: while we are in the neighbourhood, I propose replacing the few lines below with:

```
if self.blockstore.read().await.get_block(...)... {
    return;
}
```

then we do not have explicitly drop the lock.

qkniep: did that now, I just don't like the long chains and resulting line breaks

src/repair.rs
qkniep: We are not only repairing finalized blocks. We also do (and may even need to) repair blocks before they are finalized. Cases when we may need to repair blocks are: (1) a block is notarized-fallback, (2) a block matches the parent-ready vote thresholds, (3) a block is finalized (even implicitly).

src/repair.rs
akhi3030: Yeah, I find them ugly too.  But I find having to use explicit `drop` riskier and harder to read.  

We could consider changing the max width in rustfmt as a workaround to help in some cases.

PR #82, opened 2025-08-09
bugfix: handle invalid first slice when reconstructing slices and blocks

### LINKED ISSUE #79: Improve checking for parent info when reconstructing slices and blocks
in `try_reconstruct_slice` and `try_reconstruct_block`, slices with index == 0 should contain parent info.  If they don't they are invalid slices that someone "malicious" sent us.  We should improve the code so that we can handle this appropriately, i.e. not panic and construct invalid state but dropping the invalid slices, etc.

akhi3030:
closes #79

### DIFF
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -258,12 +258,12 @@ mod tests {
 
     #[tokio::test]
     async fn store_simple_block() -> Result<()> {
-        let slot = Slot::new(1);
+        let slot = Slot::genesis().next();
         let (tx, rx) = mpsc::channel(100);
         let (sk, mut blockstore) = test_setup(tx);
         assert!(blockstore.slot_data(slot).is_none());
 
-        // generate two slices for slot 1
+        // generate two slices for slot picked above
         let slices = create_random_block(slot, 2);
 
         // first slice is not enough
@@ -286,7 +286,7 @@ mod tests {
 
     #[tokio::test]
     async fn out_of_order_shreds() -> Result<()> {
-        let slot = Slot::new(1);
+        let slot = Slot::genesis().next();
         let (tx, rx) = mpsc::channel(100);
         let (sk, mut blockstore) = test_setup(tx);
         assert!(blockstore.canonical_block_hash(slot).is_none());
@@ -307,7 +307,7 @@ mod tests {
 
     #[tokio::test]
     async fn just_enough_shreds() -> Result<()> {
-        let slot = Slot::new(1);
+        let slot = Slot::genesis().next();
         let (tx, rx) = mpsc::channel(100);
         let (sk, mut blockstore) = test_setup(tx);
         assert!(blockstore.canonical_block_hash(slot).is_none());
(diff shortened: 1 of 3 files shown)


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -217,19 +217,32 @@ impl BlockData {
...
-        if slice_shreds.len() < shredder::DATA_SHREDS || self.slices.contains_key(&slice) {
+        if self.slices.contains_key(&slice) {
+            return false;
+        }
+        let reconstructed_slice = match RegularShredder::deshred(slice_shreds) {
+            Ok(s) => s,
+            Err(DeshredError::NotEnoughShreds) => return false,
+            rest => {
+                warn!("deshreding failed with {rest:?}");
+                return false;
```
akhi3030, line 228: Question: in this case and the case where the parent is not present below, shouldn't we stop receiving additional shreds for this slice as we should now have sufficient evidence that we will never properly reconstruct it.

akhi3030: We could probably handle it in subsequent PRs.

qkniep: Exactly, what we can even do is treat it the same as equivocation right now, so tigger `try_skip_window()` to vote to skip the leader

qkniep (approved):

src/consensus/blockstore.rs
```diff
@@ -260,33 +256,14 @@ mod tests {
...
-            };
-            slices.push(Slice::from_parts(header, payload, None));
-        }
-        slices
-    }
-
     #[tokio::test]
     async fn store_simple_block() -> Result<()> {
-        let slot = Slot::new(0);
+        let slot = Slot::new(1);
```
akhi3030: These and the other slots were updated because create_random_block sets the parent to slot=0.

src/consensus/cert.rs
```diff
@@ -26,7 +26,7 @@ pub enum CertError {
 }
 
 /// Certificate types used for the consensus protocol.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
akhi3030, line 29: The additional derives were added here and in other structs to support comparing VotorEvent to implement assert_votor_event_match

src/crypto/aggsig.rs
```diff
@@ -76,13 +76,13 @@ impl PublicKey {
 /// An individual signature as part of the aggregate signature scheme.
 ///
 /// This is a wrapper around [`blst::min_sig::Signature`].
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
qkniep, line 83: This only really makes sense because BLS signatures happen to be deterministic and unique. Maybe add a "NOTE" comment that these traits should be removed in case we change the underlying signature scheme?

src/consensus/blockstore.rs
qkniep: Maybe this should be `Slot::genesis().next()` and `create_random_block()` should use `Slot::genesis()`, or `create_random_block()` could take a parent as argument?

qkniep (commented):
lgtm, only two minor comments

PR #84, opened 2025-08-10
wip

akhi3030:


src/test_utils.rs
```diff
@@ -124,3 +126,24 @@ pub fn assert_votor_events_match(ev0: VotorEvent, ev1: VotorEvent) {
         }
     }
 }
+
+/// Creates a valid [`SlicePayload`] which contains valid transactions that can  be decoded.
+///
+/// HACK: Packs manually picked number of maximally sized transactions in the slice that results in creating the largest slice possible without going over the [`MAX_DATA_PER_SLICE`] limit.
```
qkniep: docstrings are usually not parsed for these comment types
```suggestion
// HACK: Packs manually picked number of maximally sized transactions in the slice that results in creating the largest slice possible without going over the [`MAX_DATA_PER_SLICE`] limit.
```

src/test_utils.rs
```diff
@@ -124,3 +126,24 @@ pub fn assert_votor_events_match(ev0: VotorEvent, ev1: VotorEvent) {
...
+fn create_random_slice_payload_valid_txs(parent: Option<BlockId>) -> SlicePayload {
+    let mut txs = vec![];
+    for _ in 0..63 {
+        let mut data = vec![0; MAX_TRANSACTION_SIZE];
```
qkniep: can be moved outside the loop

src/slice.rs
```diff
@@ -164,11 +169,12 @@ pub fn create_random_slice_payload(
 
 /// Create a [`Slice`] with a random payload of desired size.
```
qkniep:
```suggestion
/// Creates a [`Slice`] with a random payload of desired size.
```

PR #86, opened 2025-08-11
wip

akhi3030:


src/consensus/produce_block.rs
```diff
@@ -116,28 +117,58 @@ where
...
             "producing block in slot {} with parent {} in slot {} (parent ready: {})",
             slot,
             &hex::encode(parent_hash)[..8],
             parent_slot,
-            parent_ready,
+            parent_ready.is_none(),
         );
 
         let mut sleep_duration = DELTA_BLOCK;
+        let mut parent_changed = None;
```
akhi3030: bug: not handling scenario where we see a parent ready later but it is for the same parent that we have building on top of so far.  We should not produce a slice with a new parent set then.

src/consensus/produce_block.rs
```diff
@@ -113,63 +114,112 @@ where
...
+                            parent_changed = Some((new_slot, new_hash));
+                        }
+                        future.await
+                  }
                 }
+            };
+            if let Continue::Stop = &cont_prod
+                && let Some((parent_slot, parent_hash)) = parent_changed
+            {
+                assert!(slice.parent.is_none());
```
akhi3030: this is probably a bug if there is just one slice in the block.

src/consensus/produce_block.rs
```diff
@@ -113,63 +114,112 @@ where
...
+                    res = &mut receiver => {
+                        let (new_slot, new_hash) = res.unwrap();
+                        if new_slot != parent_slot && new_hash != parent_hash {
+                            parent_changed = Some((new_slot, new_hash));
+                        }
+                        future.await
+                  }
                 }
+            };
+            if let Continue::Stop = &cont_prod
```
akhi3030: - Also delta block should start only after we see parent ready
- introduce max slices per block to start accounting for CUs.
- in current version: need to ensure that parent ready has fired here.

src/consensus/produce_block.rs
```diff
@@ -185,7 +238,7 @@ where
 mod tests {
     use super::*;
 
-    use crate::{Transaction, network::UdpNetwork};
+    use crate::{Transaction, crypto::Hash, network::UdpNetwork};
```
qkniep:
```suggestion
    use crate::crypto::Hash;
    use crate::network::UdpNetwork;
    use crate::Transaction;
```

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     votor_handle: tokio::task::JoinHandle<()>,
 }
 
+enum SlotReady {
+    /// Window should be skipped.
```
qkniep:
```suggestion
    /// Window was already skipped.
```

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
...
+    Skip,
+    /// Pool emitted a parent ready event.
+    Ready(BlockId),
+    /// Canonical block for previous slot seen but pool has not emitted ParentReady yet.
```
qkniep: nit: capitalize "Pool" as a name and format `ParentReady`
```suggestion
    /// Canonical block for previous slot seen but Pool has not emitted `ParentReady` yet.
```

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     votor_handle: tokio::task::JoinHandle<()>,
 }
 
+enum SlotReady {
+    /// Window should be skipped.
+    Skip,
+    /// Pool emitted a parent ready event.
```
qkniep:
```suggestion
    /// Pool emitted a `ParentReady` for given `BlockId`.
```

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
...
+}
+
 /// Waits for first slot in the given window to become ready for block production.
 ///
 /// Ready here can mean:
 /// - Pool emitted the `ParentReady` event for it, OR
-/// - the blockstore has stored a canonical block for the previous slot.
+/// - the blockstore has stored a block for the previous slot.
 ///
 /// If the slot became ready, returns `Some((parent_slot, parent_hash, is_parent-ready))`.
```
qkniep: this docstring needs updating

src/consensus.rs
```diff
@@ -344,37 +357,39 @@ where
                 continue;
             }
 
-            let (parent, parent_hash, mut parent_ready) = match wait_for_first_slot(
+            // produce first block.
```
qkniep:
```suggestion
            // produce first block
```

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
     R: Network + Sync + Send + 'static,
     T: Network + Sync + Send + 'static,
 {
-    pub(crate) async fn produce_block(
+    async fn shred_and_disseminate(&self, slice: Slice) -> Result<()> {
```
qkniep: so far I have followed the pattern of having `pub` functions at the top of each `impl` block

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
+        mut receiver: oneshot::Receiver<BlockId>,
     ) -> Result<()> {
-        let (parent_slot, parent_hash) = parent;
         let _slot_span = Span::enter_with_local_parent(format!("slot {slot}"));
+        let (parent_slot, parent_hash) = parent_block_id;
+        assert_eq!(parent_slot, slot.prev());
+        assert!(slot.is_start_of_window());
         info!(
-            "producing block in slot {} with parent {} in slot {} (parent ready: {})",
+            "produce_block_parent_not_ready slot {} with parent {} in slot {}",
```
qkniep: so far I have used natural language for log messages, explicit function names can be automatically added in the log format if needed
```suggestion
            "optimistically producing block in slot {} with parent {} in slot {}",
```

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
+
+    pub(crate) async fn produce_block_parent_ready(
+        &self,
+        slot: Slot,
+        parent_block_id: BlockId,
+    ) -> Result<()> {
+        let _slot_span = Span::enter_with_local_parent(format!("slot {slot}"));
+        let (parent_slot, parent_hash) = parent_block_id;
+        info!(
+            "produce_block_parent_ready slot {} with parent {} in slot {}",
```
qkniep: so far I have used natural language for log messages, explicit function names can be automatically added in the log format if needed
```suggestion
            "producing block in slot {} with ready parent {} in slot {}",
```

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
             &hex::encode(parent_hash)[..8],
             parent_slot,
-            parent_ready,
         );
 
         let mut sleep_duration = DELTA_BLOCK;
-        for slice_index in 0.. {
-            let slice_index = match NonZeroUsize::new(slice_index) {
-                None => Either::Left((parent_slot, parent_hash, 0)),
+        for ind in 0.. {
```
qkniep: why not go for the more standard and no less descriptive `i`?

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
...
     }
 
     let last_slot_in_window = first_slot_in_window.last_slot_in_window();
 
     // if already have parent ready, return it, otherwise get a channel to await on
-    let rx = {
+    let mut rx = {
         let mut guard = pool.write().await;
         match guard.wait_for_parent_ready(first_slot_in_window) {
             Either::Left((slot, hash)) => {
```
qkniep: nit: could pass `BlockId` without deconstructing here

src/consensus.rs
```diff
@@ -88,44 +88,54 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
...
     // - `ParentReady` event,
-    // - canonical block reconstruction in blockstore, OR
+    // - block reconstruction in blockstore, OR
     // - notification that a later slot was finalized.
     tokio::select! {
-        res = rx => {
+        res = &mut rx => {
             let (slot, hash) = res.expect("Sender dropped channel.");
-            Some((slot, hash, true))
+            SlotReady::Ready((slot, hash))
```
qkniep: nit: could pass `BlockId` without deconstructing here

src/consensus.rs
```diff
@@ -143,7 +153,7 @@ async fn wait_for_first_slot(
                             &hex::encode(hash)[..8],
                             last_slot_in_prev_window,
                         );
```
qkniep: we already have a higher priority `info` log for this in `produce_block_no_parent_ready()`

src/consensus.rs
```diff
@@ -344,37 +357,39 @@ where
...
-            let mut block = parent;
-            let mut block_hash = parent_hash;
-            for slot in first_slot_in_window.slots_in_window() {
-                if slot.is_genesis() {
-                    parent_ready = true;
-                    continue;
+                SlotReady::Skip => continue,
+                SlotReady::Ready(parent) => {
+                    let () = self
+                        .produce_block_parent_ready(first_slot_in_window, parent)
```
qkniep: I don't think this handles genesis correctly, does it? It would call `produce_block_parent_ready` for slot 0 and would produce a block for that slot with genesis (also in slot 0) as parent, which after #64 is no longer the intended behavior. Instead, the first block produced by the block production loop should be for slot 1. We should probably have a unit test for that.

src/consensus.rs
```diff
@@ -344,37 +357,39 @@ where
...
+            let parent_slot = first_slot_in_window;
+            let parent_hash = self
+                .blockstore
+                .read()
+                .await
+                .canonical_block_hash(parent_slot)
+                .unwrap();
+            for slot in first_slot_in_window.slots_in_window().skip(1) {
+                let () = self
+                    .produce_block_parent_ready(slot, (parent_slot, parent_hash))
```
qkniep: here `parent_slot` and `parent_hash` stay the same for the entire loop, which is wrong

qkniep: instead we probably want `produce_block_parent_ready` to return the `BlockId` of the produced block (or at least the `Hash`) and assign that

qkniep: for symmetry we could also have `produce_block_no_parent_ready` return the same and change lines 382ff accordingly

src/consensus/produce_block.rs
```diff
@@ -7,15 +7,15 @@ use std::time::{Duration, Instant};
 use color_eyre::Result;
 use either::Either;
 use fastrace::Span;
-use log::{info, warn};
+use log::info;
 use static_assertions::const_assert;
+use tokio::{pin, sync::oneshot};
```
qkniep: nit: module-level grouping of imports
```suggestion
use tokio::sync::oneshot;
use tokio::pin;
```

src/consensus/produce_block.rs
```diff
@@ -27,7 +27,7 @@ enum Continue {
 async fn produce_slice<T>(
```
qkniep: also within this file I would move this private function further down, below the public parts in the `impl` blocks

src/consensus/produce_block.rs
```diff
@@ -66,19 +64,16 @@ where
...
-                    let tx = bincode::serde::encode_to_vec(&tx, bincode::config::standard())
-                        .expect("serialization should not panic");
-                    slice_capacity_left = slice_capacity_left.checked_sub(tx.len()).unwrap();
-                    txs.push(tx);
-                }
-                msg => {
-                    panic!("Unexpected msg: {msg:?}");
-                }
-            },
+        match res.expect("Unexpected error") {
```
qkniep: nit: capitalization
```suggestion
        match res.expect("unexpected error") {
```

src/consensus/produce_block.rs
```diff
@@ -66,19 +64,16 @@ where
...
+            NetworkMessage::Transaction(tx) => {
+                let tx = bincode::serde::encode_to_vec(&tx, bincode::config::standard())
+                    .expect("serialization should not panic");
+                slice_capacity_left = slice_capacity_left.checked_sub(tx.len()).unwrap();
+                txs.push(tx);
+            }
+            msg => {
+                panic!("Unexpected msg: {msg:?}");
```
qkniep: nit: capitalization
```suggestion
                panic!("unexpected msg: {msg:?}");
```

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
     R: Network + Sync + Send + 'static,
     T: Network + Sync + Send + 'static,
 {
-    pub(crate) async fn produce_block(
+    async fn shred_and_disseminate(&self, slice: Slice) -> Result<()> {
+        let slice_slot = slice.slot;
```
qkniep: unnecessary assignment, only used once

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
+                .await;
+            if let Ok(Some((slot, block_info))) = block {
+                assert_eq!(slot, slice_slot);
+                self.pool.write().await.add_block(slot, block_info).await;
+            }
+        }
+        Ok(())
+    }
+
+    /// Produces a block in the situation where we have not yet seen the Parent-Ready event.
```
qkniep:
```suggestion
    /// Produces a block in the situation where we have not yet seen the `ParentReady` event.
```

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
-                let block = self
-                    .blockstore
-                    .write()
-                    .await
-                    .add_shred_from_disseminator(s)
-                    .await;
-                if let Ok(Some((slot, block_info))) = block {
-                    self.pool.write().await.add_block(slot, block_info).await;
+
+            let produce_slice_future =
```
qkniep, line 64: maybe a comment explaining this block (lines 158-174) would be good?

src/consensus/produce_block.rs
```diff
@@ -113,63 +108,121 @@ where
...
+        let (parent_slot, parent_hash) = parent_block_id;
+        info!(
+            "produce_block_parent_ready slot {} with parent {} in slot {}",
+            slot,
+            &hex::encode(parent_hash)[..8],
+            parent_slot,
+        );
 
+        let mut sleep_duration = DELTA_BLOCK;
+        for ind in 0.. {
```
qkniep: why not go for the more standard and no less descriptive `i`?

qkniep (changes requested):

src/consensus/produce_block.rs
akhi3030: I wonder if there is some rustfmt setting that can manage the imports in the way we want.  Generally, I am letting my IDE make these changes and it seems not to be doing them in the right way.

src/consensus/produce_block.rs
akhi3030: We consume the slice in the next line.  that's why we need this assignment here.

src/consensus/produce_block.rs
akhi3030: It is a habit I developed some time ago when I used to write a lot of C and didn't have good IDE support.  And searching for `ind` was easier than searching `i`.  It is still a bit helpful when we are reading code on github (and don't have IDE support there) but in general, nowadays, I should go back to `i`.

I'll change to `i`.

src/consensus.rs
akhi3030: good catch, such a silly bug.

src/consensus/produce_block.rs
```diff
@@ -24,10 +25,160 @@ enum Continue {
...
+            self.shred_and_disseminate(slice).await?;
+            match cont_prod {
+                Continue::Stop => break,
+                Continue::Continue {
+                    left: left_duration,
+                } => sleep_duration = left_duration,
+            }
+        }
+
+        let hash = self
```
akhi3030: I'm not very happy with this code and the same in the function above.  It'll be nice for `shred_and_disseminate` to return the blockid when we hand over the final slice.  I am thinking of adding some more type safety around first and last slice.  I'll investigate in a separate PR.

src/consensus.rs
akhi3030: Does it maybe make sense to define the genesis window as containing just 1 slot and then we can simply skip the whole window?

qkniep: As far as I can tell, the setup we have now is exactly as on Solana mainnet, which is nice. It also has the slight advantage that leader windows are perfectly aligned with slot numbers.
What we can do here, is just pass `first_slot_in_window + 1` to `produce_block_parent_ready` in case `first_slot_in_window` is the genesis slot.

akhi3030: should be fixed.

src/consensus/produce_block.rs
akhi3030: I explored this idea a bit in https://github.com/qkniep/alpenglow/pull/94 but decided that it is not probably not worth it.

src/consensus/produce_block.rs
```diff
@@ -100,14 +100,21 @@ where
                 } => sleep_duration = left_duration,
             }
         }
-        Ok(())
+
+        let hash = self
```
qkniep: we could have `shred_and_disseminate()` return the block hash (e.g. as `Option<Hash>` or `Option<BlockId>`) to prevent these explicit lookups in blockstore

Issue #88, opened 2025-08-13
Enabling mocking of `Pool` and `Blockstore`


akhi3030:
Context is https://github.com/qkniep/alpenglow/pull/86 and in particular https://github.com/qkniep/alpenglow/pull/86#issuecomment-3182445431.  Essentially, I want to be able to mock these objects so that I can add testing for #86 in a simpler manner.

PR #89, opened 2025-08-13
feat: introduce traits for pool and blockstore to help with mocking

akhi3030:
Also adds a trivial test to demonstrate how the objects can be mocked.

Closes https://github.com/qkniep/alpenglow/issues/88

### DIFF
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -375,13 +375,17 @@ impl RepairResponse {
 mod tests {
     use super::*;
 
+    use crate::consensus::{BlockstoreImpl, PoolImpl};
     use crate::network::simulated::{SimulatedNetwork, SimulatedNetworkCore};
     use crate::test_utils::generate_validators;
 
     use tokio::sync::mpsc::Sender;
 
-    async fn create_repair_instance() -> (Sender<BlockId>, SimulatedNetwork, Arc<RwLock<Blockstore>>)
-    {
+    async fn create_repair_instance() -> (
+        Sender<BlockId>,
+        SimulatedNetwork,
+        Arc<RwLock<Box<dyn Blockstore + Send + Sync>>>,
+    ) {
         let (_, epoch_info) = generate_validators(2);
         let mut epoch_info = Arc::try_unwrap(epoch_info).unwrap();
         epoch_info.validators.get_mut(0).unwrap().repair_address = "0".to_string();
@@ -389,16 +393,13 @@ mod tests {
         let epoch_info = Arc::new(epoch_info);
 
         let (votor_tx, _) = tokio::sync::mpsc::channel(100);
-        let blockstore = Arc::new(RwLock::new(Blockstore::new(
-            epoch_info.clone(),
-            votor_tx.clone(),
-        )));
+        let blockstore: Arc<RwLock<Box<dyn Blockstore + Send + Sync>>> = Arc::new(RwLock::new(
+            Box::new(BlockstoreImpl::new(epoch_info.clone(), votor_tx.clone())),
+        ));
 
         let (repair_tx, repair_rx) = tokio::sync::mpsc::channel(100);
-        let pool = Arc::new(RwLock::new(Pool::new(
-            epoch_info.clone(),
-            votor_tx,
-            repair_tx.clone(),
+        let pool: Arc<RwLock<Box<dyn Pool + Send + Sync>>> = Arc::new(RwLock::new(Box::new(
+            PoolImpl::new(epoch_info.clone(), votor_tx, repair_tx.clone()),
         )));
 
         let core = Arc::new(


src/consensus.rs
```diff
@@ -97,8 +97,8 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
 /// If the slot became ready, returns `Some((parent_slot, parent_hash, is_parent-ready))`.
 /// Else returns `None` if the window should be skipped.
 async fn wait_for_first_slot(
-    pool: Arc<RwLock<Pool>>,
-    blockstore: Arc<RwLock<Blockstore>>,
+    pool: Arc<RwLock<Box<dyn Pool + Send + Sync>>>,
```
qkniep, line 109: I am a bit worried about the use of trait objects and dynamic dispatch for such central structures. Specifically, I am worried about the additional cost of dynamic dispatch on each call of `add_shred()` or `add_vote()`. Do you know if there is some cleverness of the Rust compiler to get rid of this overhead because we only have one implementor of each trait in production code?

Cargo.toml
```diff
@@ -14,6 +14,7 @@ publish = false
 [dependencies]
 aes = { version = "0.9.0-rc.0", features = ["hazmat", "zeroize"] }
 anyhow = "1"
+async-trait = "0.1.88"
```
qkniep:
```suggestion
async-trait = "0.1"
```

Cargo.toml
```diff
@@ -32,6 +33,7 @@ geo = "0.30"
 hex = "0.4"
 log = "0.4"
 logforth = { version = "0.26", features = ["colored", "append-fastrace"] }
+mockall = "0.13.1"
```
qkniep:
```suggestion
mockall = "0.13"
```

src/consensus.rs
```diff
@@ -429,3 +436,33 @@ where
...
+    use std::sync::Arc;
+
+    use tokio::sync::RwLock;
+
+    use crate::{
+        Slot,
+        consensus::{
+            Blockstore, Pool, blockstore::MockBlockstore, pool::MockPool, wait_for_first_slot,
+        },
+    };
```
qkniep:
```suggestion
    use crate::consensus::{Blockstore, Pool,  wait_for_first_slot};
    use crate::consensus::blockstore::MockBlockstore;
    use crate::consensus::pool::MockPool;
    use crate::Slot;
    
    use tokio::sync::RwLock;
    
    use std::sync::Arc;
```

src/consensus/blockstore.rs
```diff
@@ -63,6 +85,53 @@ impl Blockstore {
         }
     }
 
+    async fn send_votor_event(&self, event: VotorEvent) -> Option<(Slot, BlockInfo)> {
```
qkniep, line 98: move private method further down?

src/consensus.rs
akhi3030: Yes, if we were to use generics here like we do in `Alpenglow` then that will eliminate the extra cost.

Still, this is one the grand questions in Rust, generics or dynamic dispatch.  Dynamic dispatch incurs additional costs for lookups whereas generics have other problems. 

Some links that I looked up:
- https://users.rust-lang.org/t/how-to-avoid-passing-generics-parameters-through-the-whole-system/95750.  We are hitting some of the problem described here already.  
- https://www.linkedin.com/pulse/hidden-cost-lack-thereof-dynamic-dispatch-rust-migo-lee-qsknc.  Talks a bit about the performance savings of generics compared to the overall codebase.  In general, one can come up with lots of microbenchmarks that show savings in generics but often times, it is hard to demonstrate savings with large system performance tests.  
- https://stackoverflow.com/questions/28621980/what-are-the-actual-runtime-performance-costs-of-dynamic-dispatch.  There is also a lot of compiler work that has happened in gcc and can be done (or is already done) in rustc which can further eliminate the costs of dynamic dispatch.

I think one idea we can investigate is having 2 traits for Pool and Blockstore each.  One for a set of functions that are in the hot path and are hopefully not used in many parts of the systems.  For this trait we use generics.  And for the rest of the functions, we use dynamic dispatch.

What do you think?  We can also jump on a call to discuss next steps if that is easier.

qkniep: I think we can leave this for now, maybe open an issue to look at this again

akhi3030: @qkniep : just a ping to see if you saw this message.  It will be great to converge on next steps here as this will enable a bunch of different type of testing.

qkniep: Some of the discussions you shared are a bit reassuring regarding the overhead involved. The idea of having two different traits for each sounds very overkill to me at the moment. I propose we just keep the trait objects for now and potentially do some benchmarking regarding this later.

akhi3030: > I think we can leave this for now, maybe open an issue to look at this again

weird, I didn't see this message till now!  I think github is having some hiccups again.  Sorry for the multiple pings.

akhi3030: Created https://github.com/qkniep/alpenglow/issues/95.

qkniep: no worries, new issue looks good

PR #90, opened 2025-08-16
feat: improvements to `create_random_slice`

akhi3030:
Improves `create_random_slice` so that it actually encodes transactions in it.  Which can hopefully we decoded later.  This will be helpful in testing https://github.com/qkniep/alpenglow/pull/84.

### DIFF
--- a/src/slice.rs
+++ b/src/slice.rs
@@ -8,7 +8,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::crypto::Hash;
 use crate::shredder::{MAX_DATA_PER_SLICE, Shred};
-use crate::{Slot, highest_non_zero_byte};
+use crate::{MAX_TRANSACTION_SIZE, Slot};
 
 /// A slice is the unit of data between block and shred.
 ///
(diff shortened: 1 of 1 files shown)


akhi3030:
Marking as draft as I found some edge cases where the code currently does not work.  I'll make adjustments.

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
     }
 }
 
+/// Helper function for [`create_random_slice_payload`].
+/// Creates a Vec of [`Transaction`], binencodes it, and returns the resulting byte vector.
+/// The number of transactions used and their lens are picked so that the resulting byte vector is of size `desired_size`.
+//
+// NOTE: This function is terribly hacky and making assumptions about how bincode works.
```
qkniep, line 135: There is a perfect comment type for this:
```suggestion
// HACK: This function is terribly hacky and making assumptions about how bincode works.
```

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
...
+    txs
+}
+
 /// Creates a [`SlicePayload`] with a random payload of desired size.
 ///
 /// This function should only be used for testing and benchmarking.
 //
 // XXX: This is only used in test and benchmarking code.  Ensure it is only compiled when we are testing or benchmarking.
+//
+// NOTE: This function is terribly hacky and making assumptions about how bincode works.
```
qkniep, line 167: Same as above:
```suggestion
// HACK: This function is terribly hacky and making assumptions about how bincode works.
```

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
     }
 }
 
+/// Helper function for [`create_random_slice_payload`].
```
qkniep, line 131: nit: all docstrings should start with a one-line paragraph
```suggestion
/// Helper function for [`create_random_slice_payload`].
///
```

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
...
+/// Creates a Vec of [`Transaction`], binencodes it, and returns the resulting byte vector.
+/// The number of transactions used and their lens are picked so that the resulting byte vector is of size `desired_size`.
+//
+// NOTE: This function is terribly hacky and making assumptions about how bincode works.
+fn create_txs(desired_size: usize) -> Vec<u8> {
+    let mut rng = rng();
```
qkniep, line 137: I usually explicitly refer to `rand::rng()` to avoid confusion with the variables called `rng`.

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
...
+    let mut used = 1;
+    let mut txs = vec![];
+    while used + MAX_TRANSACTION_SIZE < desired_size {
+        let mut tx = vec![0; MAX_TRANSACTION_SIZE - 3];
```
qkniep, line 141: maybe add a comment what the "magic numbers" (3 and 6) come from, or even define them as `const`s?

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
     }
 }
 
+/// Helper function for [`create_random_slice_payload`].
+/// Creates a Vec of [`Transaction`], binencodes it, and returns the resulting byte vector.
+/// The number of transactions used and their lens are picked so that the resulting byte vector is of size `desired_size`.
+//
+// NOTE: This function is terribly hacky and making assumptions about how bincode works.
+fn create_txs(desired_size: usize) -> Vec<u8> {
```
qkniep, line 136: in general, so far I have followed the convention of ordering functions/methods from high-level (`pub`) to low-level (private), so here it would be `create_random_slice()` first, then `create_random_slice_payload()`, finally `create_txs()` at the end.

src/slice.rs
```diff
@@ -128,38 +128,52 @@ impl From<Vec<u8>> for SlicePayload {
...
-
-    payload.into()
+    let parent_payload =
+        bincode::serde::encode_to_vec(parent, bincode::config::standard()).unwrap();
+    let used = parent_payload.len();
+    let txs = create_txs(desired_size - used - 3);
+    SlicePayload::new(parent, txs)
 }
 
 /// Create a [`Slice`] with a random payload of desired size.
```
qkniep, line 179:
```suggestion
/// Creates a [`Slice`] with a random payload of desired size.
```

src/slice.rs
```diff
@@ -176,3 +190,23 @@ pub fn create_random_slice(desired_size: usize) -> Slice {
     };
     Slice::from_parts(header, payload, None)
```
qkniep, line 191: this currently produces an invalid slice with `slice_index: 0` and no parent

PR #92, opened 2025-08-18
ci: add `rustfmt`

### LINKED ISSUE #91: Enable `rustfmt` on CI
We are currently manually enforcing various formatting styles when we review PRs.  This is errorprone and adds additional complexity for the reviewer.  

We should enable `rustfmt` on CI so that these become more streamlined.  

https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=import#group_imports shows what the supported styles are grouping imports.  The style we are currently trying to achieve does not quite match what is supported.  In particular, we cannot group `super` and `crate` separately.  I propose that we give up on grouping them separately so that we can use the `StdExternalCrate` option.

qkniep:
Closes #91.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -2,4 +2,6 @@ edition = "2024"
 unstable_features = true
 
 group_imports = "StdExternalCrate"
-imports_granularity = "module"
+imports_granularity = "Module"
+reorder_impl_items = true
+use_field_init_shorthand = true


akhi3030:
In the ideal world, I would like a setup that my IDE can follow, that we can validate on the CI, and requires no or minimal manual work.

If I understand the proposal correctly, if I later start using `abc::Foo` somewhere in the code, then the IDE will have to figure out where to put the import statement (above or below the `mod abc` statement) and the CI will have to validate it.  If we are not able to enforce ordering there, then things can diverge over time.  Do you think rustfmt can help us enforce ordering?

A meta level suggestion is essentially to not enforce any grouping and put all the import statements together.

While I understand some of the benefits of grouping, in practice, I tend not to look at the actual import statements.  If I am curious about which module something is defined in, then usually I rely on the IDE support.  Having said that, I also realise that this is a bad habit for when one does not have IDE support (e.g. browsing code on github) then they have to look at the top of the files and having things be organised is a huge benefit then.

In my previous project, we ultimately gave up on grouping because we were unable to enforce it on CI (e.g. [this](https://github.com/near/nearcore/blob/01e0b7f4902f5790fadd1ebee3afc17279299f2c/core/primitives/src/transaction.rs#L222)).

PR #96, opened 2025-08-19
wip

akhi3030:


qkniep:
Blockstore should also consider `MAX_SLICES_PER_BLOCK` and never store any slice with a higher index. Should this be part of this PR?

src/consensus/produce_block.rs
```diff
@@ -280,64 +250,23 @@ mod tests {
...
-                assert_eq!(parent, Some(parent_block_id));
-                assert_eq!(data.len(), 1);
-            }
-        }
+        let parent = None;
+        let (payload, maybe_duration) =
+            produce_slice_payload(&txs_receiver, parent, duration_left).await;
+        assert!(maybe_duration.is_none());
+        assert_eq!(payload.parent, parent);
+        // bin encoding an empty vec takes 1 byte.
```
qkniep:
```suggestion
        // bin encoding an empty Vec takes 1 byte
```

src/consensus/produce_block.rs
```diff
@@ -280,64 +250,23 @@ mod tests {
...
+        assert_eq!(payload.data.len(), 1);
+
+        let parent = Some((Slot::genesis(), Hash::default()));
+        let (payload, maybe_duration) =
+            produce_slice_payload(&txs_receiver, parent, duration_left).await;
+        assert!(maybe_duration.is_none());
+        assert_eq!(payload.parent, parent);
+        // bin encoding an empty vec takes 1 byte.
```
qkniep:
```suggestion
        // bin encoding an empty Vec takes 1 byte
```

src/consensus/produce_block.rs
```diff
@@ -280,64 +250,23 @@ mod tests {
...
-                } = slice;
-                assert_eq!(slot, slot);
-                assert_eq!(slice_index, slice_index);
-                assert!(is_last);
-                assert!(merkle_root.is_none());
-                assert!(parent.is_none());
-                assert_eq!(data.len(), 1);
-            }
-        }
+        let duration_left = Duration::from_micros(1);
```
qkniep: can/should we pass actual 0 duration?

src/consensus/produce_block.rs
```diff
@@ -126,45 +112,46 @@ where
...
     ///
-    /// Returns Ok(Some(Hash)) if slice.is_last == true.
-    /// Returns Ok(None) if slice.is_last == false.
-    async fn shred_and_disseminate(&self, slice: Slice) -> Result<Option<Hash>> {
+    /// Returns Ok(Either::Left(hash of the block)) if this is the last slice.
+    /// Returns Ok(Either::Right(duration left in block)) if this is not the last slice.
+    async fn shred_and_disseminate(
+        &self,
+        slice_index: usize,
+        slot: Slot,
```
qkniep: nit: I would switch `slot` and `slice_index`, as that's the order we always use (from larger unit to smaller, e.g. epoch, slot, slice, shred)

src/consensus/produce_block.rs
```diff
@@ -126,45 +112,46 @@ where
...
+        payload: SlicePayload,
+        maybe_duration: Option<Duration>,
+    ) -> Result<Either<Hash, Duration>> {
+        let is_last = slice_index == MAX_SLICES_PER_BLOCK - 1 || maybe_duration.is_none();
+        let header = SliceHeader {
+            is_last,
+            slot,
+            slice_index,
+        };
+        let slice = Slice::from_parts(header, payload, None);
```
qkniep, line 159: why do we construct the slice from payload here now, instead of passing it to this function? seems slightly more complicated if anything (more arguments and we have to determine `is_last` in two places)

src/consensus/produce_block.rs
```diff
@@ -126,45 +112,46 @@ where
...
     }
 
-    /// Shreds and disseminates the slice.
+    /// Shreds and disseminates the slice payload.
     ///
-    /// Returns Ok(Some(Hash)) if slice.is_last == true.
-    /// Returns Ok(None) if slice.is_last == false.
-    async fn shred_and_disseminate(&self, slice: Slice) -> Result<Option<Hash>> {
+    /// Returns Ok(Either::Left(hash of the block)) if this is the last slice.
+    /// Returns Ok(Either::Right(duration left in block)) if this is not the last slice.
```
qkniep:
```suggestion
    /// Returns Ok(Either::Right(duration left in slot)) if this is not the last slice.
```

src/consensus/produce_block.rs
akhi3030: You are right, my original thinking was that I can reduce code duplication however, then I had to check for `is_last` in the other place.

I think I will still pass a SliceHeader instead of Slice to this function.  And we could then change `shred()` to maybe take the header and payload separately.  No need to construct the slice just to deconstruct it right away.

src/consensus/produce_block.rs
akhi3030: I suppose based on the comment below, this will become obsolete.

src/consensus/produce_block.rs
akhi3030: good point, I don't see why not.

PR #98, opened 2025-08-19
test: another simple test for `wait_for_first_slot`

akhi3030:
This is mainly me trying to improve my understanding of mockall.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -89,6 +89,7 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
 }
 
 /// Enum to capture the different scenarios that can be returned from [`wait_for_first_slot`].
+#[derive(Debug)]
 enum SlotReady {
     /// Window was already skipped.
     Skip,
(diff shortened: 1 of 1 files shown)


src/consensus.rs
```diff
@@ -462,4 +465,26 @@ mod tests {
             SlotReady::Ready(_)
         ));
     }
+
+    #[tokio::test]
+    async fn wait_for_first_slot_parent_already_ready() {
+        let blockstore: Box<dyn Blockstore + Send + Sync> = Box::new(MockBlockstore::new());
+        let blockstore = Arc::new(RwLock::new(blockstore));
+
+        let slot = Slot::windows().skip(10).next().unwrap();
```
qkniep: clippy suggestion:
```suggestion
        let slot = Slot::windows().nth(10).unwrap();
```

src/consensus.rs
```diff
@@ -462,4 +465,26 @@ mod tests {
...
+        let parent = (slot.prev(), Hash::default());
```
qkniep, line 475: should this not be this?
```suggestion
        let parent = (Slot::genesis(), Hash::default());
```

akhi3030: We can pick whatever parent we want here really.  The result we get back in line 486 will be the same that we provide in line 480.  The actual implementation of the Pool doesn't matter.

qkniep: ah, ok... so the `returning()` call specifies what `wait_for_parent_read()` should return when called with `slot` later?

akhi3030: Yup.  The mock stuff is pretty powerful and has a lot of knobs to control what it should return each time, etc.

qkniep (approved):
looks good then

PR #99, opened 2025-08-19
bug: emit `ParentReady` for ancestor of finalized block 

### LINKED ISSUE #97: Issue `ParentReady` upon finalizing block
Analogous to https://github.com/anza-xyz/alpenglow/pull/404.

qkniep:
When seeing a new finalized block, Pool should check whether we issued a `ParentReady` for this window already. If not a `ParentReady` should be issued for the first slot of the window.

This ensures that we correctly enter the window when catching up based on finalized blocks, setting the timeouts and voting notar or skip for the slots. Especially during standstill this is critical. Without this, if enough validators are in this state, the network may not ever recover from standstill.

Closes #97.

### DIFF
--- a/src/consensus/pool.rs
+++ b/src/consensus/pool.rs
@@ -486,7 +486,7 @@ impl Pool for PoolImpl {
 
     /// Gives the currently highest finalized (fast or slow) slot.
     fn finalized_slot(&self) -> Slot {
-        self.finality_tracker.highest_finalized()
+        self.finality_tracker.highest_finalized_slot()
     }
 
     /// Returns all possible parents for the given slot that are ready.
--- a/src/consensus/pool/finality_tracker.rs
+++ b/src/consensus/pool/finality_tracker.rs
@@ -32,16 +32,21 @@ pub struct FinalityTracker {
     /// Maps blocks to their parents.
     parents: BTreeMap<BlockId, BlockId>,
     /// The highest finalized slot so far.
-    highest_finalized: Slot,
+    highest_finalized_slot: Slot,
 }
 
 /// Possible states a slot can be in regarding finality.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum FinalizationStatus {
+    /// Block with given hash is notarized, but slot is not yet (known to be) finalized.
     Notarized(Hash),
-    Finalized,
-    FinalizedAndNotarized(Hash),
+    /// Slot is finalized, but notarized block is not yet known.
+    FinalizedSlot,
+    /// Slot is finalized, and notarized block is known to have the given hash.
+    FinalizedBlock(Hash),
+    /// Block with given hash was implicitly finalized through later finalization.
     ImplicitlyFinalized(Hash),
+    /// Slot was implicitly skipped through later finalization.
     ImplicitlySkipped,
 }
 
@@ -79,7 +84,7 @@ impl FinalityTracker {
             return FinalizationEvent::default();
         };
         match status {
-            FinalizationStatus::FinalizedAndNotarized(hash)
+            FinalizationStatus::FinalizedBlock(hash)
             | FinalizationStatus::ImplicitlyFinalized(hash) => {
                 let mut event = FinalizationEvent::default();
                 if &block_hash == hash {
@@ -88,7 +93,7 @@ impl FinalityTracker {
                 event
             }
             FinalizationStatus::Notarized(_)
-            | FinalizationStatus::Finalized
+            | FinalizationStatus::FinalizedSlot
             | FinalizationStatus::ImplicitlySkipped => FinalizationEvent::default(),
         }
     }
@@ -101,10 +106,10 @@ impl FinalityTracker {
     pub fn mark_fast_finalized(&mut self, slot: Slot, block_hash: Hash) -> FinalizationEvent {
         let old = self
             .status
-            .insert(slot, FinalizationStatus::FinalizedAndNotarized(block_hash));
+            .insert(slot, FinalizationStatus::FinalizedBlock(block_hash));
         if let Some(status) = old {
             match status {
-                FinalizationStatus::FinalizedAndNotarized(hash)
+                FinalizationStatus::FinalizedBlock(hash)
                 | FinalizationStatus::ImplicitlyFinalized(hash) => {
                     assert_eq!(hash, block_hash, "consensus safety violation");
                     return FinalizationEvent::default();
@@ -112,14 +117,14 @@ impl FinalityTracker {
                 FinalizationStatus::Notarized(hash) => {
                     assert_eq!(hash, block_hash, "consensus safety violation");
                 }
-                FinalizationStatus::Finalized => {}
+                FinalizationStatus::FinalizedSlot => {}
                 FinalizationStatus::ImplicitlySkipped => unreachable!("consensus safety violation"),
             }
         };
 
         let mut event = FinalizationEvent::default();
-        self.handle_finalized((slot, block_hash), &mut event);
-        self.highest_finalized = slot.max(self.highest_finalized);
+        self.handle_finalized_block((slot, block_hash), &mut event);
+        self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
         event
     }
 
@@ -136,17 +141,17 @@ impl FinalityTracker {
         if let Some(status) = old {
             match status {
                 FinalizationStatus::Notarized(hash)
-                | FinalizationStatus::FinalizedAndNotarized(hash)
+                | FinalizationStatus::FinalizedBlock(hash)
                 | FinalizationStatus::ImplicitlyFinalized(hash) => {
                     assert_eq!(hash, block_hash, "consensus safety violation");
                     FinalizationEvent::default()
                 }
                 FinalizationStatus::ImplicitlySkipped => FinalizationEvent::default(),
-                FinalizationStatus::Finalized => {
+                FinalizationStatus::FinalizedSlot => {
                     let mut event = FinalizationEvent::default();
                     self.status
-                        .insert(slot, FinalizationStatus::FinalizedAndNotarized(block_hash));
-                    self.handle_finalized((slot, block_hash), &mut event);
+                        .insert(slot, FinalizationStatus::FinalizedBlock(block_hash));
+                    self.handle_finalized_block((slot, block_hash), &mut event);
                     event
                 }
             }
@@ -162,24 +167,24 @@ impl FinalityTracker {
     ///
     /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
     pub fn mark_finalized(&mut self, slot: Slot) -> FinalizationEvent {
-        let old = self.status.insert(slot, FinalizationStatus::Finalized);
+        let old = self.status.insert(slot, FinalizationStatus::FinalizedSlot);
         if let Some(status) = old {
             match status {
-                FinalizationStatus::Finalized
-                | FinalizationStatus::FinalizedAndNotarized(_)
+                FinalizationStatus::FinalizedSlot
+                | FinalizationStatus::FinalizedBlock(_)
                 | FinalizationStatus::ImplicitlyFinalized(_) => FinalizationEvent::default(),
                 FinalizationStatus::Notarized(block_hash) => {
                     let mut event = FinalizationEvent::default();
                     self.status
-                        .insert(slot, FinalizationStatus::FinalizedAndNotarized(block_hash));
-                    self.handle_finalized((slot, block_hash), &mut event);
-                    self.highest_finalized = slot.max(self.highest_finalized);
+                        .insert(slot, FinalizationStatus::FinalizedBlock(block_hash));
+                    self.handle_finalized_block((slot, block_hash), &mut event);
+                    self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
                     event
                 }
                 FinalizationStatus::ImplicitlySkipped => unreachable!("consensus safety violation"),
             }
         } else {
-            self.highest_finalized = slot.max(self.highest_finalized);
+            self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
             FinalizationEvent::default()
         }
     }
@@ -187,8 +192,8 @@ impl FinalityTracker {
     /// Returns the highest finalized slot.
     ///
     /// Does not consider whether we have the block, or even know the hash.
-    pub fn highest_finalized(&self) -> Slot {
-        self.highest_finalized
+    pub fn highest_finalized_slot(&self) -> Slot {
+        self.highest_finalized_slot
     }
 
     /// Handles the direct finalization of the given block.
@@ -199,7 +204,7 @@ impl FinalityTracker {
     /// - The finalized block,
     /// - any potentially implicitly finalized blocks, AND
     /// - any implicitly skipped slots.
-    fn handle_finalized(&mut self, finalized: BlockId, event: &mut FinalizationEvent) {
+    fn handle_finalized_block(&mut self, finalized: BlockId, event: &mut FinalizationEvent) {
         event.finalized = Some(finalized);
         if let Some(parent) = self.parents.get(&finalized) {
             self.handle_implicitly_finalized(finalized.0, *parent, event);
@@ -235,8 +240,8 @@ impl FinalityTracker {
                         return;
                     }
                     FinalizationStatus::Notarized(_) => {}
-                    FinalizationStatus::Finalized
-                    | FinalizationStatus::FinalizedAndNotarized(_)
+                    FinalizationStatus::FinalizedSlot
+                    | FinalizationStatus::FinalizedBlock(_)
                     | FinalizationStatus::ImplicitlyFinalized(_) => {
                         unreachable!("consensus safety violation")
                     }
@@ -252,13 +257,13 @@ impl FinalityTracker {
             .insert(slot, FinalizationStatus::ImplicitlyFinalized(block_hash));
         if let Some(status) = old {
             match status {
-                FinalizationStatus::FinalizedAndNotarized(hash)
+                FinalizationStatus::FinalizedBlock(hash)
                 | FinalizationStatus::ImplicitlyFinalized(hash) => {
                     assert_eq!(hash, block_hash, "consensus safety violation");
                     self.status.insert(slot, status);
                     return;
                 }
-                FinalizationStatus::Notarized(_) | FinalizationStatus::Finalized => {}
+                FinalizationStatus::Notarized(_) | FinalizationStatus::FinalizedSlot => {}
                 FinalizationStatus::ImplicitlySkipped => {
                     unreachable!("consensus safety violation")
                 }


qkniep:
Added `mark_finalized()` function to the `ParentReadyTracker` now. Internally, it just marks the parent of the finalized block as notarized-fallback and all slots between finalized block and its parent as skipped. It then emits all the new `ParentReady`s. Instead this could just emit a single `ParentReady` (for the window of the finalized slot). This should not make a big difference unless there is a long stretch of skipped leaders before the finalized slot.

qkniep:
I Implemented a `FinalityTracker`, which is a very basic data structure that is now used inside pool to keep track of when a block is (1) available, (2) notarized, (3) finalized (previously we were only tracking the highest finalized slot, without caring about notarization and block being available). The `FinalityTracker` then returns the finalized block ID and its parent's ID, which can then be passed to the `ParentReadyTracker`, for issuing the potentially missing `ParentReady`.

@akhi3030 It's still ugly and I would need to add testing but I would be interested in what you think about this approach, specifically looking at the last commit.

akhi3030:
To make sure I understand the problem correctly, let's say that the window consists of slots 10, 11, 12, 13.  I join the network at slot 12 and missed the parent ready for slot 10.  I then see the finalized block for slot 12.  So now we want to issue a parent ready for 10 so that I can setup the timers for the later slots.  Is that roughly correct?

Question: doesn't need to be handled in this PR but do we need to adjust how we set the timer for the remaining slots?  i.e. if set the timers as normally, in the example above, the timer for slot 13 would fire much later.  Not sure what kind of challenges that can create.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+impl Default for FinalityTracker {
+    fn default() -> Self {
+        Self {
+            parents: BTreeMap::new(),
+            notarized: BTreeMap::new(),
+            finalized: BTreeSet::new(),
+            highest_finalized: Slot::genesis(),
```
akhi3030: maybe we can define a default for `Slot` to return genesis, then we can just derive default for this struct.

qkniep:
Implementing the last test above, I just realized that I'm still missing a fundamental part (which also appears in your example), where the first block of the window may only be finalized implicitly. Thus we don't just need (1) the finalized block, (2) its notar cert, (3) its final cert, but also (4) any ancestor blocks in the same window.

> Is that roughly correct?

Yes, in that case we would want to issue ParentRead(10, slot(b), hash(b)) where block b is the ancestor of the block in slot 12 with the highest slot number < 10.

> Can you expand a bit on what do you not like about the `FinalityTracker`?

I think what I currently don't like is that it seem to be so simple, that there might be a nice way to do entirely without it and just have some additional stuff in `Pool`. However, thinking about it maybe that's not a good idea: Pool is already complicated enough and especially when also considering the logic that is still missing (see first paragraph above) `FinalityTracker` may become more non-trivial as well.

> do we need to adjust how we set the timer for the remaining slots?

Trying to fire the timeouts for individual slots early seems like over-engineering to me: This should very rarely happen. If it happens, we are already entering the window late, for all we know we could have been offline for minutes, trying to set timeouts exactly right seems like a lost cause.

However, if we really wanted to do something, it should certainly be safe to deduct `DELTA_BLOCK` for each block up to and including the finalized block in the window from all timeouts for the window. So, in your example, we could deduct `3 * DELTA_BLOCK`.

akhi3030:
makes sense.  Thanks for the additional context.  I agree that at the moment `FinalityTracker` does seem quite simple.  Let's see what it looks like once you have added the additional case of implicit finalization.

qkniep:
Not entirely sure if my introduction of an `enum FinalizationStatus` in the `FinalityTracker` made things more or less understandable. But now it should also support implicit finalization of blocks and implicit skipping of slots. And now I would say `FinalityTracker` is no longer trivial, it actually became more complex than expected.

src/consensus/pool/parent_ready_tracker/parent_ready_state.rs
```diff
@@ -86,12 +95,21 @@ impl ParentReadyState {
...
-            IsReady::Ready(block_ids) => Either::Left(*block_ids.first().unwrap()),
+            IsReady::Ready(block_ids) => {
+                assert!(!block_ids.is_empty());
+                if block_ids.len() == 1 {
+                    Either::Left(block_ids[0])
+                } else {
+                    let mut block_ids = block_ids.to_vec();
+                    block_ids.sort();
+                    Either::Left(block_ids[0])
+                }
```
akhi3030:
```suggestion
                block_ids.sort();
                Either::Left(block_ids[0])
```

This seems to cargo check.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct FinalizationEvent {
+    /// Directly finalized block, if any.
+    pub(super) finalized: Option<BlockId>,
+    /// Any implicitly finalized blocks.
+    pub(super) implicitly_finalized: Vec<BlockId>,
+    /// Any implicitly skipped slots.
+    pub(super) implicitly_skipped: Vec<Slot>,
+}
+
+impl FinalityTracker {
```
akhi3030, line 64: nit: I prefer to group the struct definitions and impls together.  Not a big deal though.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Adds the given `parent` for the given `block`.
+    ///
+    /// Handles possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn add_parent(&mut self, block: BlockId, parent: BlockId) -> FinalizationEvent {
```
akhi3030, line 70: nit: I would be worried that someone might accidentally switch block and parent around.  Could we add some debug asserts (e.g. parent.slot < block.slot)?

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+use crate::crypto::Hash;
+use crate::types::Slot;
+
+/// Tracks finality of blocks.
+#[derive(Default)]
+pub struct FinalityTracker {
+    /// Current finalization status for each slot.
+    status: BTreeMap<Slot, FinalizationStatus>,
+    /// Maps blocks to their parents.
+    parents: BTreeMap<BlockId, BlockId>,
```
akhi3030, line 33: probably future work.  Thinking about https://github.com/qkniep/alpenglow/pull/99/files#r2290476466, I wonder if we should use a new type for BTreeMap here which will validate all inserts.  Might be an overkill though.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// Handles possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn add_parent(&mut self, block: BlockId, parent: BlockId) -> FinalizationEvent {
+        let mut event = FinalizationEvent::default();
+        self.parents.insert(block, parent);
+        if let Some(status) = self.status.get(&block.0)
+            && let FinalizationStatus::FinalizedAndNotarized(block_hash)
+            | FinalizationStatus::ImplicitlyFinalized(block_hash) = status
+            && block.1 == *block_hash
```
akhi3030:
```suggestion
            && &block.1 == block_hash
```

nit.  I think the compiler will probably produce the same code in both cases.  But it seems better like this as at least the code claims that we are not copying.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+        {
+            self.handle_implicitly_finalized(block.0, parent, &mut event);
```
akhi3030: I struggled a bit with this function:

- the tuple indexing was difficult to think about.
- The if statement took some time to parse
- The if statement does not handle other variants and if we modify the enum, we might miss something.

So I came up with the following.  A bit more verbose but hopefully more readable.

```suggestion
        let (slot, block_hash) = block;
        if let Some(status) = self.status.get(&slot) {
            match status {
                FinalizationStatus::FinalizedAndNotarized(hash)
                | FinalizationStatus::ImplicitlyFinalized(hash) => {
                    if &block_hash == hash {
                        self.handle_implicitly_finalized(block.0, parent, &mut event);
                    }
                }

                FinalizationStatus::Notarized(_)
                | FinalizationStatus::Finalized
                | FinalizationStatus::ImplicitlySkipped => (),
            }

```

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+    ///
+    /// If the block was newly finalized, handles resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn mark_fast_finalized(&mut self, slot: Slot, block_hash: Hash) -> FinalizationEvent {
+        let mut event = FinalizationEvent::default();
+        if let Some(status) = self.status.get(&slot) {
+            match status {
+                FinalizationStatus::FinalizedAndNotarized(_) => {
+                    return event;
```
akhi3030: nit: I would just return `FinalizationEvent::default()` here and define event after the if statement instead of on the top.  The way the code is right now, it seems like we may modify event in the if statement and use the modifications below.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl FinalityTracker {
+    /// Adds the given `parent` for the given `block`.
+    ///
+    /// Handles possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn add_parent(&mut self, block: BlockId, parent: BlockId) -> FinalizationEvent {
+        let mut event = FinalizationEvent::default();
+        self.parents.insert(block, parent);
```
akhi3030: What are the implications of if the block already had a parent?

src/consensus/pool/finality_tracker.rs
akhi3030: What are the implications of the block hash not matching?

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+            match status {
+                FinalizationStatus::Notarized(_)
+                | FinalizationStatus::FinalizedAndNotarized(_)
+                | FinalizationStatus::ImplicitlyFinalized(_)
+                | FinalizationStatus::ImplicitlySkipped => event,
+                FinalizationStatus::Finalized => {
+                    self.status
+                        .insert(slot, FinalizationStatus::FinalizedAndNotarized(block_hash));
+                    self.handle_finalized((slot, block_hash), &mut event);
+                    event
```
akhi3030: same comment as above.  I would define the event above this line and in the rest of the arms return `FinalizationEvent::default()`

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,305 @@
...
+    /// - any potentially implicitly finalized blocks, AND
+    /// - any implicitly skipped slots.
+    fn handle_finalized(&mut self, finalized: BlockId, event: &mut FinalizationEvent) {
+        event.finalized = Some(finalized);
+        if let Some(parent) = self.parents.get(&finalized) {
+            self.handle_implicitly_finalized(finalized.0, *parent, event);
+        }
+    }
+
+    /// Handles the indirect finalization of the given block.
```
akhi3030, line 214: If the node joined the network after being absent for a long time, this function could recurse a lot.  There might be a potential for running out of stack space.  Probably not something we need to worry about in this PR but could be something to track in an issue at least.

akhi3030 (commented):
I didn't review the whole patch as I felt like the number of comments was getting a bit out of hand.  

They are a bit overlapping so let me know if some of them don't make sense.

src/consensus/pool/finality_tracker.rs
qkniep: good suggestion, added the same assert for `Pool::add_block()`

src/consensus/pool/finality_tracker.rs
qkniep: `add_parent()` is idempotent if the same parent is added twice because `handle_implicit_finalization` is idempotent, a different parent should never be added, we could add an assertion for that

qkniep: added an assertion and early return now, thus making idempotence of `add_parent()` very explicit

src/consensus/pool/finality_tracker.rs
qkniep: it just means we added a parent relation between block b and parent(b), whereas the block that was finalized is b' instead, so we don't care about b and its parent... this can happen, for example, if we just received block b from Rotor, but also already saw notar+final certs for b'

src/consensus/pool/finality_tracker.rs
qkniep: seems overkill to me, `FinalityTracker` is already just a glorified wrapper around the two `BTreeMap`s

src/consensus/pool/finality_tracker.rs
qkniep: I don't think we need to worry about stack space, the compiler should be able to optimize this tail recursion, otherwise we could also write it as a loop explicitly. Even when catching up, we would only trigger implicit finalizations at most as quickly as we repair the ancestor chain of the directly finalized block. The only case where this should ever trigger a massive chain, is if for some reason we *first* repair a long chain of blocks and only then see the finalization for the last block in the chain.

src/consensus/pool/finality_tracker.rs
qkniep: I went for something slightly different but hopefully just as readable

src/consensus/pool/finality_tracker.rs
qkniep: so far we are mostly consistent in preferring the other style, I have no problem with moving to the struct+impl style, but this should be a separate refactoring PR then

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// Handles possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn add_parent(&mut self, block: BlockId, parent: BlockId) -> FinalizationEvent {
+        assert!(block.0 > parent.0);
+        if let Some(p) = self.parents.get(&block) {
+            assert!(p == &parent);
+            return FinalizationEvent::default();
+        }
+        self.parents.insert(block, parent);
```
akhi3030: We are doing two accesses into the map here.  We could either check the return value from `insert` or use the `entry` API.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    /// Marks the given block as notarized.
+    ///
+    /// Handles possibly resulting direct finalization of the block.
+    /// Further, also handles any possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn mark_notarized(&mut self, slot: Slot, block_hash: Hash) -> FinalizationEvent {
+        if let Some(status) = self.status.get(&slot) {
```
akhi3030: we are accessing the map a couple of different times in this function.  I wonder if the entries api will be more readable.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,344 @@
...
+            match status {
+                FinalizationStatus::Notarized(_)
+                | FinalizationStatus::FinalizedAndNotarized(_)
+                | FinalizationStatus::ImplicitlyFinalized(_)
+                | FinalizationStatus::ImplicitlySkipped => FinalizationEvent::default(),
+                FinalizationStatus::Finalized => {
```
akhi3030: I think when we first marked the slot as finalized, we called `handle_finalized` there.  So calling it again here should strictly speaking not be necessary?  Or I am missing something.

src/consensus/pool/finality_tracker.rs
```diff
@@ -0,0 +1,344 @@
...
+    ///
+    /// Handles possibly resulting direct finalization of a block in this slot.
+    /// Further, also handles any possibly resulting implicit finalizations.
+    ///
+    /// Returns a [`FinalizationEvent`] that contains information about newly finalized slots.
+    pub fn mark_finalized(&mut self, slot: Slot) -> FinalizationEvent {
+        if let Some(status) = self.status.get(&slot) {
+            match status {
+                FinalizationStatus::Finalized
+                | FinalizationStatus::FinalizedAndNotarized(_)
```
akhi3030: We can get multiple finalizations for a slot if we got fast finalization and slow one?  Are there other scenarios?

akhi3030 (commented):
A partial review of just the finality_tracker.  I will review the changes in the pool either tonight or tomorrow.

src/consensus/pool/finality_tracker.rs
qkniep: This should be correct as is, the idea is: We want to call `handle_finalized()` only after the block is notarized and the slot is finalized. Notarization and finalization can happen in any order, so in `mark_notarized()` should only call it if the slot is already finalized, and `mark_finalized()` only if a block in the slot was notarized before.

It might be possible to make this more clear in a few places. Do you think referring to finalized slot and finalized block would be clearer, e.g. `FinalizationStatus::FinalizedSlot` and `FinalizationStatus::FinalizedBlock(hash)` and `handle_finalized_block()`.

src/consensus/pool/finality_tracker.rs
qkniep: changed in a bunch of places, see https://github.com/qkniep/alpenglow/pull/99/commits/321657f5e01190e8426b1ee9feb07bd1b96e1698

src/consensus/pool/finality_tracker.rs
qkniep: yes, if `FinalityTracker` is used correctly there should be at most two triggers (or even just one) to finalize a block in a given slot, however there are potentially many ways finalization can be triggered within pool (constructing FastFinal cert ourselves, receiving it, constructing Notar+Final cert ourselves, receiving those, and then there are many different certs) most of which should be ignored if they come after the first one

I think low-level components like `FinalityTracker` (and other parts of the code) ideally should not break if these events are triggered more than once by some higher-level component

src/consensus/pool/finality_tracker.rs
akhi3030: >  Do you think referring to finalized slot and finalized block would be clearer, e.g. FinalizationStatus::FinalizedSlot and FinalizationStatus::FinalizedBlock(hash) and handle_finalized_block().


Yes, I think that can definitely help!

src/consensus/pool/finality_tracker.rs
akhi3030: that makes sense.  I agree that these components should be resilient to such events as well.

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -77,10 +64,11 @@ impl ParentReadyTracker {
     /// Marks the given slot as skipped.
     ///
     /// Returns a list of any newly connected parents.
-    ///
-    /// This should only ever be called once for any specific slot.
     pub fn mark_skipped(&mut self, slot: Slot) -> Vec<(Slot, BlockId)> {
```
akhi3030, line 67: The changes really improve the readability of the function.  Thank you.

src/consensus/pool.rs
```diff
@@ -141,10 +144,16 @@ impl PoolImpl {
...
-                self.highest_notarized_fallback_slot =
-                    slot.max(self.highest_notarized_fallback_slot);
+                info!(
+                    "notarized(-fallback) block {} in slot {}",
+                    &hex::encode(block_hash)[..8],
+                    slot
+                );
+                if matches!(cert, Cert::Notar(_)) {
+                    let finalization_event = self.finality_tracker.mark_notarized(slot, block_hash);
+                    self.parent_ready_tracker
```
akhi3030: to help future readers, should we add a comment about why we can ignore the returned parent ready here.

qkniep: good catch, that was actually a bug! wrapped the two calls in a `handle_finalization()` function to prevent this

PR #101, opened 2025-08-19
wip

akhi3030:


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -151,8 +152,9 @@ impl BlockData {
         // store last slice index, delete everything after last slice
         if is_last_slice && self.last_slice.is_none() {
             self.last_slice = Some(slice_index);
-            self.slices.split_off(&(slice_index + 1));
-            self.shreds.split_off(&(slice_index + 1));
+            let next_index = SliceIndex::new(slice_index.inner() + 1).unwrap();
```
akhi3030: needs fixing.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -250,7 +252,7 @@ impl BlockData {
             return None;
         }
         let last_slice = self.last_slice?;
-        if self.slices.len() != last_slice + 1 {
+        if self.slices.len() != last_slice.inner() + 1 {
```
akhi3030, line 253: needs fixing.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -302,8 +304,8 @@ impl BlockData {
         self.completed = Some((block_hash, block));
 
         // clean up raw slices
-        for slice_index in 0..=last_slice {
-            self.slices.remove(&slice_index);
+        for slice_index in 0..=last_slice.inner() {
```
akhi3030: needs fixing.

src/repair.rs
```diff
@@ -153,7 +154,7 @@ impl<N: Network> Repair<N> {
                 RepairResponse::SliceCount(request, k)
             }
             RepairRequest::SliceRoot(_, _, slice) => {
-                let shred_index = slice * TOTAL_SHREDS;
+                let shred_index = slice.inner() * TOTAL_SHREDS;
```
akhi3030, line 157: needs fixing.

src/shredder.rs
```diff
@@ -174,8 +174,8 @@ pub struct ShredPayload {
 impl ShredPayload {
     /// Returns the index of this shred within the entire slot.
     #[must_use]
-    pub const fn index_in_slot(&self) -> usize {
-        self.header.slice_index * DATA_SHREDS + self.index_in_slice
+    pub fn index_in_slot(&self) -> usize {
+        self.header.slice_index.inner() * DATA_SHREDS + self.index_in_slice
```
akhi3030, line 178: needs fixing

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -151,8 +152,8 @@ impl BlockData {
         // store last slice index, delete everything after last slice
         if is_last_slice && self.last_slice.is_none() {
             self.last_slice = Some(slice_index);
-            self.slices.split_off(&(slice_index + 1));
-            self.shreds.split_off(&(slice_index + 1));
+            self.slices.retain(|&ind, _| ind <= slice_index);
```
akhi3030, line 154: we could also keep comparing the last index and popping.  Not sure what is more efficient.

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
+use serde::{Deserialize, Serialize};
```
qkniep:
```suggestion
// Copyright (c) Anza Technology, Inc.
// SPDX-License-Identifier: Apache-2.0

use serde::{Deserialize, Serialize};
```

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl SliceIndex {
+    /// Creates a new SliceIndex for testing purposes.
+    #[cfg(test)]
+    pub(crate) fn new(index: usize) -> Self {
+        assert!(index < MAX_SLICES_PER_BLOCK);
+        Self(index)
+    }
+
+    /// Returns the inner usize.
```
qkniep:
```suggestion
    /// Returns the inner `usize`.
```

src/test_utils.rs
```diff
@@ -66,7 +67,7 @@ pub fn create_random_block(slot: Slot, num_slices: usize) -> Vec<Slice> {
         let payload = create_random_slice_payload_valid_txs(parent);
         let header = SliceHeader {
             slot,
-            slice_index,
+            slice_index: SliceIndex::new(slice_index),
```
qkniep: we don't even need to use this here, right? we can just use `until()` here as well

qkniep: and then also use `slice_index.is_first()` above (in line 62)

src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod network;
 pub mod repair;
 pub mod shredder;
 pub mod slice;
+pub mod slice_index;
```
qkniep: maybe this should just be a sub-module of `slice`?

qkniep: or we have a different module that contains `slice`, `slot`, and `slice_index`

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
+use serde::{Deserialize, Serialize};
+
+const MAX_SLICES_PER_BLOCK: usize = 1024;
+
+/// Slice index type.
+///
+/// Using strong type to enforce certain constraints e.g. it is never >= MAX_SLICES_PER_BLOCK.
```
qkniep:
```suggestion
/// Using strong type to enforce certain constraints, e.g. it is never >= MAX_SLICES_PER_BLOCK.
```

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
...
+        assert!(index < MAX_SLICES_PER_BLOCK);
+        Self(index)
+    }
+
+    /// Returns the inner usize.
+    pub(crate) fn inner(self) -> usize {
+        self.0
+    }
+
+    /// Returns the first i.e. smallest SliceIndex.
```
qkniep:
```suggestion
    /// Returns the first, i.e. smallest, `SliceIndex`.
```

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) fn first() -> Self {
+        Self(0)
+    }
+
+    /// Returns true if self is the first i.e. the smallest SliceIndex.
```
qkniep:
```suggestion
    /// Returns `true` if self is the first, i.e. smallest, `SliceIndex`.
```

src/slice_index.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) fn is_first(&self) -> bool {
+        self.0 == 0
+    }
+
+    /// Returns true if self is the last i.e. MAX_SLICES_PER_BLOCK - 1 SliceIndex.
```
qkniep:
```suggestion
    /// Returns `true` if self is the last, i.e. `MAX_SLICES_PER_BLOCK - 1`, SliceIndex.
```

src/consensus/blockstore/slot_block_data.rs
qkniep: mh, too bad we can't use `split_off()` anymore, it would be O(log n), as opposed to O(n)...

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -228,15 +229,15 @@ impl BlockData {
                 return false;
             }
         };
-        if reconstructed_slice.parent.is_none() && reconstructed_slice.slice_index == 0 {
+        if reconstructed_slice.parent.is_none() && reconstructed_slice.slice_index.is_first() {
             warn!(
-                "reconstructed slice {} in slot {} expected to contain parent",
+                "reconstructed slice {:?} in slot {} expected to contain parent",
```
qkniep: `SliceIndex` should implement `Display`

qkniep (changes requested):
We still do not check the slice index for shreds that we receive, right? We should either check during deserialization that the `usize` we are deserializing is in the correct range (probably preferable), or we should explicitly check during `add_shred_XXX()` functions.

akhi3030:
> We still do not check the slice index for shreds that we receive, right? We should either check during deserialization that the `usize` we are deserializing is in the correct range (probably preferable), or we should explicitly check during `add_shred_XXX()` functions.

Great catch.  I had made a mental note that I cannot derive deserilize and need to implement it myself to implement the bounds checking but then forgot.

src/lib.rs
akhi3030: We can put the three structs in the same file but I would advocate for different modules i.e. using `mod` to separate them.  I had to deal with two pretty nasty bugs that would've been avoided if we had sufficient information hiding.  Essentially, some developer accidentally accessed a member field that they shouldn't have and the compiler didn't help because they were in the same module.

Let me know what you prefer and I am happy to move things around.

qkniep: I am fine with them being in different files as well as different modules, just think they should not all be at the top level.

akhi3030: In the past, we had a top level `types` module that contained some of the "basic types".  We could make these three submodules of types?

src/test_utils.rs
akhi3030: fixed this and a few other occurrences as well.

src/lib.rs
akhi3030: done.

src/network/simulated.rs
```diff
@@ -265,12 +268,13 @@ mod tests {
         let mut rng = rand::rng();
         let sk = SecretKey::new(&mut rng);
         let mut shreds = Vec::new();
-        for i in 0..10_000 {
+        let final_slice_index = SliceIndex::new_unchecked(9_999);
```
akhi3030: @qkniep: I just realized that this test will fail now because we are capping max slice index at 1024.  I suppose I can introduce a way to create a slice index that does not assert that it is < 1024 and use that here.  Or should we consider rewriting the test to use multiple slots, or any other ideas?

qkniep: this test also works with fewer slices, I just reduced the number of slices to 1024 now

PR #103, opened 2025-08-20
metrics proposal from the research team

akhi3030:
This pseudocode tries to capture some epoch level metrics that the research team would like to have.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -1,0 +1,102 @@
+use std::{collections::BTreeMap, time::Duration};
+
+use histogram::Histogram;
+
+use crate::{ValidatorId, consensus::vote::VoteKind};
+
+struct VoteMetrics {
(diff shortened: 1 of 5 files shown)


src/metrics.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl VoteMetrics {
+    fn new() -> Self {
+        let histogram = Histogram::new(10, 10).unwrap();
+        Self {
+            notar: (0, histogram.clone()),
+            notar_fallback: (0, histogram.clone()),
+            skip: (0, histogram.clone()),
+            skip_fallback: (0, histogram.clone()),
+            final_: (0, histogram),
```
rogerANZA, line 23: why no clone?

akhi3030:
closing as the proposal has now landed in agave.

PR #104, opened 2025-08-20
refactor: move block production into a separate struct

akhi3030:
This should make it easier to be able to create test setups for various methods in the new `BlockProducer` struct.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -69,6 +69,7 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     /// Pool of votes and certificates.
     pool: Arc<RwLock<Box<dyn Pool + Send + Sync>>>,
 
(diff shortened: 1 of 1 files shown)


src/consensus.rs
```diff
@@ -73,103 +69,21 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     /// Pool of votes and certificates.
     pool: Arc<RwLock<Box<dyn Pool + Send + Sync>>>,
 
+    block_producer: Arc<BlockProducer<D, T>>,
```
qkniep, line 73:
```suggestion
    /// Block production (i.e. leader side) component of the consensus protocol.
    block_producer: Arc<BlockProducer<D, T>>,
```

qkniep (approved):
nice, LGTM

PR #106, opened 2025-08-21
feat: perform actual block repair

### LINKED ISSUE #12: Perform actual block repair
As of now, there is only one pseudo-repair function for requesting the parent info of a specific block without proof.

Instead we should implement a minimal version of the actual repair protocol, which should:
- Repair the actual block data.
- Include proofs (e.g. Merkle paths) in the responses.
- Request different parts in parallel.

~This requires #16 to be completed first.~

qkniep:
Closes #12.

So far repair only requests the parent information. We then ingest the parent information into `Pool`. This is enough for consensus to continue. However, it bypasses the main point of repairing the actual block. Also, currently the `Parent` responses do not contain any proof that the parent is the actual parent of the requested block. In this PR we replace this with actual repair of block, where each response contains a proof.

### DIFF
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -9,7 +9,7 @@
 //! Each repair response is accompanied by a Merkle proof and can thus be
 //! individually verified.
 
-use std::collections::{BTreeMap, BinaryHeap};
+use std::collections::{BTreeMap, BinaryHeap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -359,8 +359,14 @@ impl<N: Network> Repair<N> {
         let msg: NetworkMessage = repair.into();
         let msg_bytes = msg.to_bytes();
         // HACK: magic number to fix high-failure scenarios
-        for _ in 0..3 {
-            let to = self.pick_random_peer();
+        let mut to_all = HashSet::new();
+        for _ in 0..10 {
+            to_all.insert(self.pick_random_peer());
+            if to_all.len() == 3 {
+                break;
+            }
+        }
+        for to in to_all {
             self.network.send_serialized(&msg_bytes, to).await?;
         }
         Ok(())
@@ -418,7 +424,6 @@ mod tests {
     use crate::consensus::{BlockstoreImpl, PoolImpl};
     use crate::crypto::signature::SecretKey;
     use crate::network::simulated::{SimulatedNetwork, SimulatedNetworkCore};
-    use crate::shredder::DATA_SHREDS;
     use crate::test_utils::{create_random_shredded_block, generate_validators};
     use crate::types::Slot;
     use crate::types::slice_index::MAX_SLICES_PER_BLOCK;
@@ -547,9 +552,9 @@ mod tests {
 
             // expect Shred requests for this slice next
             let mut shreds_requested = BTreeSet::new();
-            for _ in 0..DATA_SHREDS {
+            for _ in 0..TOTAL_SHREDS {
                 let msg = other_network.receive().await.unwrap();
-                for shred_index in 0..DATA_SHREDS {
+                for shred_index in 0..TOTAL_SHREDS {
                     let req = RepairRequest::Shred(block_to_repair, slice, shred_index);
                     if msg_matches_request(&msg, &req) {
                         shreds_requested.insert(shred_index);
@@ -560,7 +565,7 @@ mod tests {
 
             // assert all shreds requested + answer the requests
             let slice_shreds = shreds[slice.inner()].clone();
-            for (shred_index, shred) in slice_shreds.into_iter().take(DATA_SHREDS).enumerate() {
+            for (shred_index, shred) in slice_shreds.into_iter().take(TOTAL_SHREDS).enumerate() {
                 assert!(shreds_requested.contains(&shred_index));
                 let req = RepairRequest::Shred(block_to_repair, slice, shred_index);
                 let response = RepairResponse::Shred(req, shred);
@@ -639,7 +644,7 @@ mod tests {
             assert_eq!(proof, correct_proof);
 
             // request slice shreds
-            for shred_index in 0..DATA_SHREDS {
+            for shred_index in 0..TOTAL_SHREDS {
                 let request = RepairRequest::Shred(block_to_repair, slice, shred_index);
                 let msg = RepairMessage::Request(request.clone(), 0).into();
                 other_network.send(&msg, "1").await.unwrap();


qkniep:
Sorry for another PR that is getting (too) big. I now removed the old `Parent` repair request and response types. In the future, we could think about reintroducing this (with Merkle proofs) in addition to regular repair. This would probably make repair-related quite quite a bit more complicated but could speed up certain recovery scenarios, e.g., when we only need to know the parent hash for to cast safe-to-notar.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -90,7 +90,7 @@ impl SlotBlockData {
             .entry(hash)
             .or_insert_with(|| BlockData::new(self.slot));
         match block_data.check_shred_to_add(&shred, true, leader_pk) {
-            Ok(()) => Ok(self.canonical.add_valid_shred(shred)),
+            Ok(()) => Ok(block_data.add_valid_shred(shred)),
```
akhi3030, line 93: as discussed on slack.  We should have a test to ensure that we are not accidentally using `self.canonical` here.

akhi3030, line 93: we also discussed renaming `self.canonical` to something else.  Leaving a note here so that we do not forget.

src/consensus/blockstore.rs
```diff
@@ -53,12 +54,19 @@ pub trait Blockstore {
         shred: Shred,
     ) -> Result<Option<(Slot, BlockInfo)>, AddShredError>;
     fn canonical_block_hash(&self, slot: Slot) -> Option<Hash>;
-    fn stored_shreds_for_slot(&self, slot: Slot) -> usize;
     #[allow(clippy::needless_lifetimes)]
-    fn get_block<'a>(&'a self, slot: Slot, hash: Hash) -> Option<&'a Block>;
+    fn get_block<'a>(&'a self, block_id: BlockId) -> Option<&'a Block>;
+    fn get_last_slice(&self, block_id: BlockId) -> Option<SliceIndex>;
```
akhi3030:
```suggestion
    fn get_last_slice_index(&self, block_id: BlockId) -> Option<SliceIndex>;
```

The name seems to suggest that we will return the actual slice but we are returning the index.

src/consensus/blockstore.rs
```diff
@@ -53,12 +54,19 @@ pub trait Blockstore {
...
+    fn get_slice_root(&self, block_id: BlockId, slice: SliceIndex) -> Option<Hash>;
```
akhi3030, line 60:
```suggestion
    fn get_slice_root(&self, block_id: BlockId, index: SliceIndex) -> Option<Hash>;
```

Or `slice_index`.  `slice` suggests that we are passing in actual slice.  It is not a problem here but when reading code that calls this function, the IDE will provide the name of the parameter and if it says `slice` that makes one think that it is the `Slice` type

src/consensus/blockstore.rs
```diff
@@ -53,12 +54,19 @@ pub trait Blockstore {
...
     #[allow(clippy::needless_lifetimes)]
-    fn get_shred<'a>(&'a self, slot: Slot, slice: SliceIndex, shred: usize) -> Option<&'a Shred>;
-    fn create_double_merkle_proof(&self, slot: Slot, slice: SliceIndex) -> Vec<Hash>;
+    fn get_shred<'a>(
+        &'a self,
+        block_id: BlockId,
+        slice: SliceIndex,
+        shred: usize,
```
akhi3030:
```suggestion
        slice_index: SliceIndex,
        shred_index: usize,
```

Same reasoning as above.

src/consensus/blockstore.rs
```diff
@@ -53,12 +54,19 @@ pub trait Blockstore {
...
+    ) -> Option<&'a Shred>;
+    fn create_double_merkle_proof(&self, block_id: BlockId, slice: SliceIndex)
```
akhi3030:
```suggestion
    fn create_double_merkle_proof(&self, block_id: BlockId, index: SliceIndex)
```

src/consensus/blockstore.rs
```diff
@@ -203,51 +255,53 @@ impl Blockstore for BlockstoreImpl {
...
-    /// Gives the number of stored shreds for a given `slot` (across all slices).
-    fn stored_shreds_for_slot(&self, slot: Slot) -> usize {
-        self.slot_data(slot)
-            .map_or(0, |s| s.canonical.shreds.values().map(Vec::len).sum())
+    /// Gives reference to stored block for the given `block_id`.
+    ///
+    /// Returns `None` if blockstore does not hold that block.
+    fn get_block(&self, block_id: BlockId) -> Option<&Block> {
+        let block_data = self.get_block_data(block_id)?;
+        block_data.completed.as_ref().map(|(_, block)| block)
```
akhi3030: nit: should we add a debug assert to check that block_id.hash matches the hash on the block_data?  It will help the reader understand the properties of how the data is stored mainly.

src/consensus/blockstore.rs
```diff
@@ -203,51 +255,53 @@ impl Blockstore for BlockstoreImpl {
...
     }
 
-    /// Gives reference to stored block for the given `slot` and `hash`.
+    /// Gives reference to stored block for the given `block_id`.
```
akhi3030: update?

src/crypto/merkle.rs
```diff
@@ -12,12 +12,41 @@
...
 
-const LEAF_LABEL: [u8; 22] = *b"\x00ALPENGLOW-MERKLE-TREE";
-const LEFT_LABEL: [u8; 22] = *b"\x01ALPENGLOW-MERKLE-TREE";
-const RIGHT_LABEL: [u8; 22] = *b"\x02ALPENGLOW-MERKLE-TREE";
+const LEAF_LABEL: [u8; 32] = *b"ALPENGLOW-MERKLE-TREE  LEAF-NODE";
+const LEFT_LABEL: [u8; 32] = *b"ALPENGLOW-MERKLE-TREE  LEFT-NODE";
+const RIGHT_LABEL: [u8; 32] = *b"ALPENGLOW-MERKLE-TREE RIGHT-NODE";
+
+const EMPTY_ROOTS: [Hash; 24] = [
+    hex!("50c4672b7a309041b109458b1f3a11f82c225970975a95cd1025209301fcbcab"),
```
akhi3030, line 42: should we add a comment about how these were generated and the length being 24?

src/crypto/merkle.rs
```diff
@@ -121,6 +150,33 @@ impl MerkleTree {
...
+        let hash = hash_leaf(data);
+        Self::check_hash_proof_last(hash, index, root, proof)
+    }
+
+    /// Checks a Merkle path proves the given leaf hash is last in the tree.
+    ///
+    /// Returns `true` iff the Merkle proof is valid and `index` is the last leaf in the tree.
+    #[must_use]
+    pub fn check_hash_proof_last(hash: Hash, index: usize, root: Hash, proof: &[Hash]) -> bool {
+        assert!(proof.len() <= EMPTY_ROOTS.len(),);
```
akhi3030:
```suggestion
        assert!(proof.len() <= EMPTY_ROOTS.len());
```

src/consensus.rs
```diff
@@ -79,8 +79,7 @@ pub struct Alpenglow<A: All2All, D: Disseminator, R: Network, T: Network> {
     all2all: Arc<A>,
     /// Block dissemination network protocol for shreds.
     disseminator: Arc<D>,
-    /// Block repair protocol.
-    repair: Arc<Repair<R>>,
+    _repair: std::marker::PhantomData<Repair<R>>,
```
akhi3030: it's worth leaving a comment here explaining why we need phantom data.  It is generally a subtle reason that sometimes takes a while to wrap ones head around it.

akhi3030 (commented):
mainly cosmetic feedback so far.  I will take a break and then review repair.rs

src/crypto/merkle.rs
qkniep: note to self: add const_assert(log2(max_slices) <= 24)

src/repair.rs
```diff
@@ -54,26 +57,30 @@ pub enum RepairRequest {
 /// If well-formed, it thus contains the corresponding variant of [`RepairRequest`].
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum RepairResponse {
-    /// Response with the total number of slices for a specific block.
-    SliceCount(RepairRequest, usize),
+    /// Response with the last slice's Merkle root hash, plus corresponding proof.
+    LastSliceRoot(RepairRequest, SliceIndex, Hash, Vec<Hash>),
```
akhi3030, line 62: If `RepairRequest` is not of `LastSliceRoot` variant, then this should be invalid.  I know it will be a bit of explosion of types, but introducing `LastSliceRootRequest` would help.  I think it might be a good idea to introduce `LastSliceRootReply` for keeping things equivalent.  

If you agree, we should also do this in a separate PR though.  It will not change the logic much and will only add noise to an already big PR.

src/repair.rs
```diff
@@ -9,42 +9,45 @@
...
+    Request(RepairRequest, ValidatorId),
     /// Response to a request from another validator.
     Response(RepairResponse),
 }
 
 /// Request messages for the repair sub-protocol.
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub enum RepairRequest {
     /// Request for the total number of slices in block with a given hash.
```
akhi3030, line 46: needs updating?

src/consensus/blockstore/slot_block_data.rs
qkniep: now I am very confused, after merging with main this is fixed, but on main it's still wrong???

src/repair.rs
```diff
@@ -54,26 +57,30 @@ pub enum RepairRequest {
...
+
+    slice_roots: BTreeMap<(BlockId, SliceIndex), Hash>,
+    outstanding_requests: BTreeMap<Hash, RepairRequest>,
+    request_timeouts: BinaryHeap<(Instant, Hash)>,
+
     network: N,
     repair_channel: (
         tokio::sync::mpsc::Sender<BlockId>,
-        tokio::sync::Mutex<tokio::sync::mpsc::Receiver<BlockId>>,
+        Option<tokio::sync::mpsc::Receiver<BlockId>>,
```
akhi3030, line 81: i find this a bit un-ergonomic.  But I think we can leave this is as is for this PR and explore ways of improving it subsequently.

src/consensus/blockstore/slot_block_data.rs
qkniep: any preference what to call it? I think I like `initial` and `alternatives` (maybe `others` or `repaired`)

src/repair.rs
```diff
@@ -54,26 +57,30 @@ pub enum RepairRequest {
...
 /// Instance of double-Merkle based block repair protocol.
 pub struct Repair<N: Network> {
     blockstore: Arc<RwLock<Box<dyn Blockstore + Send + Sync>>>,
     pool: Arc<RwLock<Box<dyn Pool + Send + Sync>>>,
+
+    slice_roots: BTreeMap<(BlockId, SliceIndex), Hash>,
+    outstanding_requests: BTreeMap<Hash, RepairRequest>,
+    request_timeouts: BinaryHeap<(Instant, Hash)>,
+
     network: N,
```
akhi3030, line 78: also something to consider doing in future PRs if we want to address it.

Currently, we are using the same port to handle requests and responses.  It might make sense to use separate ports for them so that we could prioritise replies over requests.

If we use separate ports, we can also have a separate struct for handling repair requests from the network.  That big of logic essentially has to look stuff up from blockstore and send replies.  Putting it in the same struct as where we repairing is not really necessary.

src/repair.rs
```diff
@@ -93,164 +100,223 @@ impl<N: Network> Repair<N> {
...
+        loop {
+            let next_timeout = self.request_timeouts.peek().map(|t| t.0);
+            let sleep_duration = match next_timeout {
+                None => std::time::Duration::MAX,
+                Some(t) => t.duration_since(Instant::now()),
+            };
+            tokio::select! {
+                // handle repair request or response from network
+                res = self.receive() => self.handle_repair_message(res.unwrap()).await.unwrap(),
+                // handle request for repairing new block
```
akhi3030, line 136:
```suggestion
                // handle request from other components in the node for repairing new block
```

Because we are inside a macro, my IDE is not giving me type hints and it is not super clear which arm is handling network messages and which one is handling messages on a channel.  

Not sure if my suggestion above helps improve things.  Maybe you have another idea.

akhi3030: Another suggestion is to put in some explicit type annotations to help the reader.

src/repair.rs
```diff
@@ -93,164 +100,223 @@ impl<N: Network> Repair<N> {
...
+            RepairMessage::Response(resposne) => {
+                self.handle_response(resposne).await;
+            }
         }
+        Ok(())
     }
 
     /// Tries to answer the given repair request.
     ///
     /// If we do not have the block the request refers to, the request is ignored.
```
akhi3030, line 182: Again, probably future work.  If we ignore requests that we cannot satisfy, the the requester will have to wait till timeout to try again.  Does it make sense to send a response so that they can retry faster?

src/repair.rs
```diff
@@ -93,164 +100,223 @@ impl<N: Network> Repair<N> {
...
-    pub async fn answer_request(&self, request: RepairRequest) -> Result<(), NetworkError> {
+    async fn answer_request(
+        &self,
+        request: RepairRequest,
+        sender: ValidatorId,
+    ) -> Result<(), NetworkError> {
         trace!("answering repair request: {request:?}");
-        let slot = request.slot();
-        let hash = request.block_hash();
+        let block_id = request.block_id();
```
akhi3030: nit: I would probably not use this and look up the block_id explicitly in the match below.  

Having code structured like this causes some pain for the reader:

- when I look at the individual request handling and see that it contains info that is ignored, I wonder why was that info included in the first place.
- Makes me wonder if I have to look into the implementation of `request.block_id()` to see if something extra has to be done to figure the block id out.

src/consensus/blockstore/slot_block_data.rs
akhi3030: I was thinking of `disseminated` and `repaired`.

akhi3030: the way it appears to me, it is this PR that is fixing the the problem.

The diff on github shows that this PR is changing the code from using `self.canonical` to `block_data`.

src/consensus.rs
qkniep: got rid of it by making `R` a type parameter on `new()` instead

src/repair.rs
```diff
@@ -93,164 +100,223 @@ impl<N: Network> Repair<N> {
...
-                let Some(block) = blockstore.get_block(slot, hash) else {
+            RepairRequest::Shred(_, slice, shred) => {
+                let Some(shred) = blockstore.get_shred(block_id, slice, shred).cloned() else {
                     return Ok(());
                 };
-                RepairResponse::Parent(request, block.parent, block.parent_hash)
+                RepairResponse::Shred(request, shred)
             }
         };
         drop(blockstore);
```
akhi3030: It'll be great to use scopes to drop the blockstore.  I always feel nervous about not dropping locks in the right place.  Scopes feel a bit safer.

src/repair.rs
qkniep: so, you would have 6 different types instead of the 2 enums (or even 2 enums + 6 structs)? note also, that we already do checks for this

src/repair.rs
qkniep: only used in tests

src/repair.rs
qkniep: yes, it's not great

src/repair.rs
qkniep: true, these two parts could definitely be separated

src/repair.rs
qkniep: that seems to be an IDE/setup specific problem, I can see types w/o problem here

src/repair.rs
qkniep: this can probably be part of #146

qkniep:
This should finally fix the breaking test case. The problem was that repair is too slow (slower than our liveness bound of 10s in these tests) in the high failure scenario with 33% crash failures and 33% having skipped the block (i.e. not having it). I introduced some workarounds and magic numbers to make repair fast and more reliable at the cost of having more repair traffic. Will improve in future PRs.

PR #111, opened 2025-08-22
feat: final bits of optimistic handover

### LINKED ISSUE #4: Implement optimistic handover

akhi3030:
Closes #4.

Final bits of optimistic handover.  The leader switches the parent and when reconstructing the block, we check for the updated parent.

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -178,13 +178,13 @@ where
                     res = &mut parent_ready_receiver => {
                         // Got ParentReady event while producing slice.
                         // It's a NOP if we have been using the same parent as before.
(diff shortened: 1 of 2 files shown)


qkniep (approved):
LGTM

PR #112, opened 2025-08-23
test: add more tests to `network` module

qkniep:
Part of #71.


src/network.rs
```diff
@@ -186,4 +186,24 @@ mod tests {
         let deserialized = NetworkMessage::from_bytes(&bytes).unwrap();
         assert!(matches!(deserialized, NetworkMessage::Pong));
     }
+
+    #[tokio::test]
+    async fn serialize_reuse_buffer() {
+        let mut buf = [0u8; MTU_BYTES];
+        for _ in 0..10 {
+            let msg = NetworkMessage::Ping;
+            let num_bytes = msg.write_bytes(&mut buf);
```
akhi3030, line 195: comment not related to this PR:

- `write_bytes` sounds like this function is actually writing the data to network somehow.  Maybe we can call it, `serialize`?  But we have a sister function `to_bytes`.  Maybe then we can call this function `to_slice`?
- I think `write_bytes` will panic if the buffer is not big enough.  Should we keep that behaviour?  I am fine with it, just wanted to highlight it and maybe we should document it.

akhi3030 (approved):

src/network.rs
qkniep: See #117

Issue #13, opened 2025-07-04
Get ready for benchmarks


qkniep:
Some steps are necessary before we can do any meaningful benchmarks:
- [x] Pack individual (dummy) transactions into blocks. (#26)

akhi3030:
Adding some notes here from a discussion earlier.

- We have a traffic generator that is configured to generate traffic at a certain rate.
- The traffic generator needs to know what the leader schedule is so that it can send the TXs to the current leader.  There will be some timing issues here which will cause some txs to not get included / dropped.  In the future, we could implement the current leader forwarding the txs to the future leaders to help with this.
- We can measure throughput at the individual validators of how many txs per block and how many blocks per sec they are seeing.
- e2e latency is a bit trickier.  But we probably do not care about e2e from the user's point of view.  We really care about the latency from the block being proposed to it being finalised.

One initial note on building this initially is that we can build everything on a single machine but we should assume a multi-machine deployment.  i.e. we need some orchestration layer to start the validators that can talk over ssh / etc. to tell other machines what binary to start; what arguments to provide; and collect some metrics from it.

I wonder if we could use some off-the-shelf technology for it e.g. terraform.  But maybe initially that might be a bit of an overkill....

PR #118, opened 2025-08-24
feat: a simple workload generator

### LINKED ISSUE #47: Provide a way to generate workloads
Before we can run insightful benchmarks using multiple machines, we need to implement a workload generator. This should probably be a standalone binary that sends transactions to the validators on the corresponding port.

qkniep:
It is it's own binary called `workload_generator` and can generate and send random transaction of a given size and at a given TPS to the validator's transaction receiver port (all provided as CLI arguments).

Some initial runs on my MacBook show that it currently maxes out at ~320k TPS (of 20 bytes each, ~50 Mbit/s) or ~200k TPS (of 512 bytes each, ~800 Mbit/s).

Closes #47.
Part of #13.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -30,7 +30,7 @@ pub fn enable_logforth_stderr() {
 }
 
 fn enable_logforth_append<A: logforth::Append>(to_append: A) {
-    let filter = EnvFilter::from_default_env_or("alpenglow=debug,info");
+    let filter = EnvFilter::from_default_env_or("info");
     logforth::builder()
         .dispatch(|d| d.filter(filter).append(to_append))
         .apply();


src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+use alpenglow::network::NetworkMessage;
+use alpenglow::{Transaction, ValidatorInfo, logging};
+use clap::Parser;
+use color_eyre::Result;
+use log::info;
+use rand::RngCore;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ConfigFile {
```
akhi3030: doesn't appear to be used anywhere?

src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    // enable fancy `color_eyre` error messages
+    color_eyre::install()?;
+
+    logging::enable_logforth();
+
+    // parse args & load config from file
```
akhi3030:
```suggestion
    // parse args
```

src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+    let args = Args::parse();
+
+    let validator_addr = args.validator;
+    let addr: SocketAddr = validator_addr.parse().unwrap();
```
akhi3030: I think we can change the type of `args.validator` to `SocketAddr`.

The following works for me with the change:

```
cargo run --bin workload_generator -- --validator 123.4.5.6:1234
```

src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    tokio::time::sleep(Duration::from_secs(args.initial_delay_secs.unwrap_or(0))).await;
+
+    let start_time = Instant::now();
+    let mut last_report = Instant::now();
+    let mut txs_sent = 0;
+
+    let mut rng = rand::rng();
+    let mut buf = vec![0; bytes_per_tx as usize];
+    socket.set_nonblocking(true).unwrap();
```
akhi3030: I would generally expect a udp socket to never block but doesn't hurt to have this.  I think we could add a comment like the following though:

// making nonblocking so if the generator is not keeping up, it fails instead of artificially slowing down

src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+    initial_delay_secs: Option<u64>,
+    /// Target throughput in transactions per second.
+    #[arg(long)]
+    transactions_per_second: Option<u64>,
+    /// Bytes per transaction.
+    #[arg(long)]
+    transaction_size: Option<u64>,
+}
+
+#[tokio::main]
```
akhi3030: I suppose this doesn't need to be async at the moment (we are just using sleep which we could do without async) but will probably in the near future when we want to produce transactions from many threads.

src/bin/workload_generator.rs
```diff
@@ -0,0 +1,99 @@
...
+        txs_sent += 1;
+
+        let elapsed = start_time.elapsed().as_secs_f64();
+        let expected_elapsed = txs_sent as f64 / target_tps as f64;
+        if elapsed < expected_elapsed {
+            let delta = expected_elapsed - elapsed;
+            tokio::time::sleep(Duration::from_secs_f64(delta)).await;
+        }
+
+        if last_report.elapsed().as_secs() > 1 {
```
akhi3030, line 74: As future work, we should think about how to collect some metrics on alpenglow to estimate how many of the txs it actually managed to put on the chain and how many got dropped.

src/bin/workload_generator.rs
qkniep: copied from `src/bin/node.rs` and forgot to delete, surprised clippy doesn't complain

src/bin/workload_generator.rs
qkniep: I think we can even explicitly ignore errors when sending, because we don't really care if any packets are dropped, as long as we count the number of messages leaving the generator correctly

src/bin/workload_generator.rs
qkniep: true, initially I was using the udp socket from our `network` module, which is `async`

src/bin/workload_generator.rs
qkniep: yes, I would imagine this to be part of #115

PR #120, opened 2025-08-24
fix: optimistic block production leading to skips

### LINKED ISSUE #119: Fix skipped leader during normal operation
When running `run.sh` right now, every second leader is skipped. This seems to be a problem with block production.

qkniep:
Optimistic block production was blocking on producing a slice with timeout duration `Duration::MAX` before the `ParentReady` event has been seen. If this slice never completed, e.g. because we don't see any transaction, we block forever and never start the blocktime timeout. The way I fixed this now, is by limiting the time per slice before `ParentReady` was seen to a maximum of `DELTA_BLOCK / 2`.

Closes #119.

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -7,7 +7,7 @@ use std::time::{Duration, Instant};
 use color_eyre::Result;
 use either::Either;
 use fastrace::Span;
-use log::{info, warn};
+use log::{debug, info, warn};
 use static_assertions::const_assert;
 use tokio::pin;
 use tokio::sync::{RwLock, oneshot};
@@ -91,12 +91,9 @@ where
             // don't do anything if we are not the leader
             let leader = self.epoch_info.leader(first_slot_in_window);
             if leader.id != self.epoch_info.own_id {
-                continue;
-            }
-
-            if self.pool.read().await.finalized_slot() >= last_slot_in_window {
-                warn!(
-                    "ignoring window {first_slot_in_window}..{last_slot_in_window} for block production"
+                debug!(
+                    "[val {}] not producing in window {first_slot_in_window}..{last_slot_in_window}, not leader",
+                    self.epoch_info.own_id
                 );
                 continue;
             }
@@ -109,7 +106,12 @@ where
             )
             .await
             {
-                SlotReady::Skip => continue,
+                SlotReady::Skip => {
+                    warn!(
+                        "skipping window {first_slot_in_window}..{last_slot_in_window} for block production"
+                    );
+                    continue;
+                }
                 SlotReady::Ready(parent) => {
                     if first_slot_in_window.is_genesis() {
                         // genesis block is already produced so skip it
@@ -411,7 +413,7 @@ async fn wait_for_first_slot(
     // - notification that a later slot was finalized.
     tokio::select! {
         res = &mut rx => {
-            let parent = res.expect("Sender dropped channel.");
+            let parent = res.expect("sender dropped channel");
             SlotReady::Ready(parent)
         }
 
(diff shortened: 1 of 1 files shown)


src/consensus/block_producer.rs
```diff
@@ -109,7 +106,12 @@ where
             )
             .await
             {
-                SlotReady::Skip => continue,
+                SlotReady::Skip => {
+                    warn!(
```
akhi3030, line 110: nit: we have a similar warning in `wait_for_first_slot` when we produce the skip variant.  Maybe we only need to keep one.

src/consensus/block_producer.rs
```diff
@@ -184,10 +189,20 @@ where
...
+                            debug!(
+                                "changed parent from {} in slot {} to {} in slot {}",
+                                &hex::encode(parent_hash)[..8],
+                                parent_slot,
+                                &hex::encode(new_hash)[..8],
+                                new_slot
+                            );
                             payload.parent = Some((new_slot, new_hash));
+                        } else {
+                            debug!("parent is ready, continuing with same parent");
```
akhi3030, line 201: I think there is another bug here that is in the original code as well.  What is the hashes are the same but the slot has changed?

src/consensus/block_producer.rs
```diff
@@ -164,8 +166,11 @@ where
             };
 
             // If we have not yet received the ParentReady event, wait for it concurrently while producing the next slice.
-            let produce_slice_future =
-                produce_slice_payload(&self.txs_receiver, parent, duration_left);
+            let produce_slice_future = produce_slice_payload(
+                &self.txs_receiver,
+                parent,
+                duration_left.min(DELTA_BLOCK / 2),
```
akhi3030: could you add a comment here to explain why delta_block / 2 is being used please?

src/consensus/block_producer.rs
```diff
@@ -420,23 +435,19 @@ async fn wait_for_first_slot(
                 // PERF: These are burning a CPU. Can we use async here?
                 loop {
                     let last_slot_in_prev_window = first_slot_in_window.prev();
-                    if let Some(hash) = blockstore
-                        .read()
-                        .await
+                    if let Some(hash) = blockstore.read().await
```
akhi3030, line 436: hmm...  I don't think you changed the content of this line.  Are our editors formatting these lines differently?  That can get annoying.  I didn't think it would be possible in rust.

src/consensus/block_producer.rs
```diff
@@ -420,23 +435,19 @@ async fn wait_for_first_slot(
...
                         .canonical_block_hash(last_slot_in_prev_window)
                     {
                         return Some((last_slot_in_prev_window, hash));
                     }
-                    if pool.read().await.finalized_slot() >= last_slot_in_window {
-                        warn!(
-                            "ignoring window {first_slot_in_window}..{last_slot_in_window} for block production"
-                        );
+                    if pool.read().await.finalized_slot() >= first_slot_in_window {
```
akhi3030, line 441: yes, this change makes sense.  

For my understanding, if the first slot in the window has been finalized somehow (probably skipped), we are assuming that the the rest of the window will also get finalized like that eventually due to liveness guarantees?

src/consensus/block_producer.rs
akhi3030: There is another TODO that we need to handle a few lines below.

I created https://github.com/qkniep/alpenglow/issues/121.  I can handle both of them in that issue.

src/consensus/block_producer.rs
akhi3030: also, please use `self.delta_block` instead of the constant to aid in testing.

src/consensus/block_producer.rs
qkniep: it can only mean two things: (1) a later block (after our window was finalize), then our entire window has already been skipped, or (2) a block in our window has been finalized, then we massively screwed up in our validator operations (e.g. by running two instances of our validator at the same time), and if we continue producing blocks we risk equivocation and thus slashing

src/consensus/block_producer.rs
qkniep: oh, I think I manually changed that to try something, and `rustfmt` doesn't apply because it's in the `tokio::select` macro

src/consensus/block_producer.rs
qkniep: same hash in different slots should be technically impossible without a hash collision, since each shred contains the slot number

src/consensus/block_producer.rs
akhi3030: ah... phew, the world makes more sense.

src/consensus/block_producer.rs
akhi3030: got it!

src/consensus/block_producer.rs
qkniep: actually set it to `delta_block` now, should also work and is more intuitive and less restrictive than `delta_block / 2`

PR #122, opened 2025-08-24
fix: correctly account for first slice in crashed leader timeout

### LINKED ISSUE #65: Crashed leader timeout buffer
Right now the crashed-leader timeout is triggered after `DELTA_TIMEOUT` has elapsed. This is correct assuming the leader would send out a slice immediately when their leader window starts.

qkniep:
This introduces the additional parameter `DELTA_FIRST_SLICE` a short timeout for producing the first slice. The leader makes sure to only take at most this amount of time for the first slice. In return, we can account for the crashed leader timeout correctly. See issue #65 for more info.

Closes #65.

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -269,6 +269,7 @@ where
             } else {
                 None
             };
+            // make sure first slice is produced on time
             let (payload, maybe_duration) = if slice_index.is_first() {
                 let time_for_slice = duration_left.min(self.delta_first_slice);
                 produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -170,7 +170,7 @@ where
                 None
             };
 
-            let time_for_slice = if slice_index == SliceIndex::first() {
+            let time_for_slice = if slice_index.is_first() {
                 // make sure first slice is produced on time
                 // TODO: this can be made more accurate, only needed if production of first slice
                 // still takes more than delta_first_slice after we saw ParentReady, not if:
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -270,12 +270,13 @@ where
                 None
             };
             // make sure first slice is produced on time
-            let (payload, maybe_duration) = if slice_index.is_first() {
-                let time_for_slice = duration_left.min(self.delta_first_slice);
-                produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await
+            let time_for_slice = if slice_index.is_first() {
+                duration_leaft.min(self.delta_first_slice)
             } else {
-                produce_slice_payload(&self.txs_receiver, parent, duration_left).await
+                duration_left
             };
+            let (payload, maybe_duration) = 
+                produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await;
             let is_last = slice_index.is_max() || maybe_duration.is_none();
             let header = SliceHeader {
                 slot,
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -271,7 +271,7 @@ where
             };
             let time_for_slice = if slice_index.is_first() {
                 // make sure first slice is produced on time
-                duration_leaft.min(self.delta_first_slice)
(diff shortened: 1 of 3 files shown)


src/consensus/block_producer.rs
```diff
@@ -165,11 +170,21 @@ where
                 None
             };
 
-            // cap timeout for each slice to `DELTA_BLOCK`
-            // makes sure optimistic block production yields before timeout would expire
-            let time_for_slice = duration_left.min(self.delta_block);
+            let time_for_slice = if slice_index == SliceIndex::first() {
```
```suggestion
            let time_for_slice = if slice_index.is_first() {
```

src/consensus/block_producer.rs
```diff
@@ -254,8 +269,12 @@ where
             } else {
                 None
             };
-            let (payload, maybe_duration) =
-                produce_slice_payload(&self.txs_receiver, parent, duration_left).await;
+            let (payload, maybe_duration) = if slice_index == SliceIndex::first() {
```
```suggestion
            let (payload, maybe_duration) = if slice_index.is_first() {
```

src/consensus/block_producer.rs
```diff
@@ -254,8 +269,12 @@ where
...
+                let time_for_slice = duration_left.min(self.delta_first_slice);
+                produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await
+            } else {
+                produce_slice_payload(&self.txs_receiver, parent, duration_left).await
+            };
```
akhi3030: Maybe something like below has less code duplication.


```suggestion
            let time_for_slice = if slice_index.is_first() {
                duration_left.min(self.delta_first_slice)
            } else {
                duration_left
            };
            let (payload, maybe_duration) = produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await;
```

src/consensus/block_producer.rs
akhi3030: Also should we duplicate the comment from above: i.e. "maybe sure first slice is produced on time".

src/consensus/block_producer.rs
```diff
@@ -165,11 +170,21 @@ where
                 None
             };
 
-            // cap timeout for each slice to `DELTA_BLOCK`
-            // makes sure optimistic block production yields before timeout would expire
-            let time_for_slice = duration_left.min(self.delta_block);
+            let time_for_slice = if slice_index == SliceIndex::first() {
+                // make sure first slice is produced on time
+                // TODO: this can be made more accurate, only needed if production of first slice
```
akhi3030, line 175: For my understanding: it can still be the case that the leader has not seen parent ready but the other nodes have seen it.  In that case, the way the code is currently would make sense, i.e. the leader should not delay producing the first slice longer.

src/consensus/votor.rs
```diff
@@ -252,12 +253,16 @@ impl<A: All2All + Sync + Send + 'static> Votor<A> {
         );
         let sender = self.event_sender.clone();
         tokio::spawn(async move {
-            tokio::time::sleep(DELTA_TIMEOUT).await;
+            tokio::time::sleep(DELTA_TIMEOUT + DELTA_FIRST_SLICE).await;
```
akhi3030, line 256: I guess we need to create a similar issue in agave to handle this scenario.  I think it is still doing it the old way.

akhi3030 (approved):
LGTM.   Minor comments.  Disabled auto-merge.

src/consensus/votor.rs
qkniep: it was part of this issue https://github.com/anza-xyz/alpenglow/issues/243

src/consensus/block_producer.rs
qkniep: not really, without the optimistic handover the leader would only even start producing the block after seeing the parent ready, the timeouts capture that

src/consensus/votor.rs
akhi3030: We might need a new issue.  https://github.com/anza-xyz/alpenglow/blob/master/votor/src/timer_manager/timers.rs#L61.  We are generating the crashed leader timeout after DELTA_BLOCK.

qkniep: immediately after DELTA_TIMEOUT, before the first DELTA_BLOCK, if I am reading that correctly, but yes, either would warrant a new issue

PR #123, opened 2025-08-25
Adds a test for and handles some edge cases for optimistic handover

### LINKED ISSUE #121: Bug and some more work for optimistic handover
- Need to handle the TODO in https://github.com/qkniep/alpenglow/blob/main/src/consensus/block_producer.rs#L200
- Need to handle the situation where the hash didn't change but the slot changed in https://github.com/qkniep/alpenglow/blob/main/src/consensus/block_producer.rs#L186

akhi3030:
closes #121 

- Adds a test case for `produce_block_parent_not_ready`
- handles a couple of left over edge cases in optimistic handover

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -186,7 +186,7 @@ where
                 produce_slice_payload(&self.txs_receiver, parent, time_for_slice);
 
             // If we have not yet received the ParentReady event, wait for it concurrently while producing the next slice.
-            let (payload, maybe_duration) = if parent_ready_receiver.is_terminated() {
+            let (mut payload, maybe_duration) = if parent_ready_receiver.is_terminated() {
                 produce_slice_future.await
             } else {
                 pin!(produce_slice_future);
@@ -204,6 +204,7 @@ where
                         let (new_slot, new_hash) = res.unwrap();
                         let (mut payload, _maybe_duration) = produce_slice_future.await;
                         if new_hash != parent_hash {
+                            assert_ne!(new_slot, parent_slot);
                             debug!(
                                 "changed parent from {} in slot {} to {} in slot {}",
                                 &hex::encode(parent_hash)[..8],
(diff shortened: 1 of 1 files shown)


src/consensus/block_producer.rs
```diff
@@ -188,7 +188,7 @@ where
                         let start = Instant::now();
                         let (new_slot, new_hash) = res.unwrap();
                         let (mut payload, _maybe_duration) = produce_slice_future.await;
-                        if new_hash != parent_hash {
+                        if new_hash != parent_hash || new_slot != parent_slot {
```
akhi3030: do we need to check for slot?

src/consensus/block_producer.rs
```diff
@@ -284,6 +294,7 @@ where
         payload: SlicePayload,
         maybe_duration: Option<Duration>,
     ) -> Result<Either<Hash, Duration>> {
+        println!("handling head = {header:?}");
```
akhi3030: remove prints

src/consensus/block_producer.rs
```diff
@@ -649,4 +662,110 @@ mod tests {
...
+        let block_info = BlockInfo {
+            hash: [1; 32],
+            parent: (slot.prev(), [2; 32]),
+        };
+
+        let (first_slice_finished_tx, first_slice_finished_rx) = oneshot::channel();
+        let (start_second_slice_tx, start_second_slice_rx) = oneshot::channel();
+
+        let mut seq = Sequence::new();
+        let mut blockstore = MockBlockstore::new();
```
qkniep, line 718: This use of the MockBlockstore looks more complicated than doing the same for the actual blockstore. Or ist there some detail in this setup that I might be missing?

src/consensus/block_producer.rs
```diff
@@ -212,9 +212,18 @@ where
             let is_last = slice_index.is_max() || maybe_duration.is_none();
             if is_last && !parent_ready_receiver.is_terminated() {
```
qkniep, line 234: This is only to handle the case where we produce all slices before seeing the ParentReady, right?

src/consensus/block_producer.rs
qkniep: I would say, no, the first parent we reconstructed (thus we know it's valid), the second parent was reconstructed by good nodes (thus we know it's valid), a valid block has different hash if the slot is different.

akhi3030: OK, yes, that makes sense.  Thanks.

src/consensus/block_producer.rs
akhi3030: Yes.

src/consensus/block_producer.rs
akhi3030: No, I agree, this block of code is too complicated.  I wanted to checkpoint this PR and attempt to reproduce the test using the actual blockstore to see what that takes.

src/consensus/block_producer.rs
qkniep: We might be able to do without this. Any rational block producer would want to prevent that from ever happening, instead waiting longer to produce their slices. Also, we might want to have a higher slice count limit for optimistic block production anyways. One generally reasonable strategy would be to wait with producing the last slice until the timeout runs out (so way after ParentReady is emitted). I will create a new issue for this.

PR #126, opened 2025-08-25
actual blockstore test for optimistic block production

qkniep:


src/test_utils.rs
```diff
@@ -36,6 +36,30 @@ pub fn generate_validators(num_validators: u64) -> (Vec<SecretKey>, Arc<EpochInf
     (voting_sks, epoch_info)
 }
 
+pub fn generate_validators_with_leader_keys(
```
qkniep, line 39: this should probably be merged with the function above, just didn't want to break all references to `generate_validators()` in all tests

src/test_utils.rs
```diff
@@ -77,6 +101,24 @@ pub fn create_random_block(slot: Slot, num_slices: usize) -> Vec<Slice> {
     slices
 }
 
+pub fn create_random_shredded_block(
```
qkniep, line 104: this is taken from #106

src/consensus/block_producer.rs
```diff
@@ -640,12 +690,43 @@ mod tests {
...
+            pool,
+            disseminator,
+            Duration::from_micros(0),
+        );
+
+        let (parent_ready_tx, parent_ready_rx) = oneshot::channel();
+
+        let sk = &sks[epoch_info.leader(slot).id as usize];
+        let (_, _, shreds) = create_random_shredded_block(slot, 2, sk);
+        let mut parent_ready_tx = Some(parent_ready_tx);
```
qkniep, line 704: didn't know how else to get this to work with the borrow checker complaining about `parent_ready_tx` being moved in the loop

PR #130, opened 2025-08-25
wip: fix crashes

### LINKED ISSUE #129: Failed assertions when running local cluster
After the most recent PR, "./run.sh" crashes with various failed assertions.

I also did one run before pulling the last merged PR, and there we were producing blocks way faster than `DELTA_BLOCK`. We should investigate if this is still the case after the crashes have been fixed.

Maybe this is a sign, that we should have at least one of the local cluster tests running in CI as well.

qkniep:
Closes #129.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -4,22 +4,10 @@ slow_tests () {
 	echo "🐌 Running slow tests can take up to 10 minutes!"
 	echo "Starting in 3 seconds..."
 	sleep 3
-	# run performance tests in release mode (others run in debug mode)
-	cargo test --release -- --test-threads=1 --ignored \
-		high_bandwidth \
-		unlimited_bandwidth \
-		turbine_sampler \
-		turbine_sampler_real_world \
-		only_correct_nodes \
-		many_nodes \
-		single_crash \
-		max_fast_crashes \
-		too_many_fast_crashes \
-		max_crashes \
-		too_many_crashes \
-		three_nodes \
-		three_nodes_crash && \
-	RUST_BACKTRACE=1 cargo test
+	# run performance tests sequentially in release mode
+	cargo nextest run --release --jobs=1 --run-ignored=only
+	# run other tests in parallel in debug mode
+	RUST_BACKTRACE=1 cargo nextest run
 }
 
 fast_tests () {
@@ -31,10 +19,12 @@ fast_tests () {
 sequential_tests () {
 	echo "Running sequential tests!"
 	sleep 1
-	RUST_BACKTRACE=1 cargo test -- test-threads=1 --ignored \
+	RUST_BACKTRACE=1 cargo nextest run --release --jobs=1 --run-ignored=only \
 		network::simulated::core::tests::asymmetric \
-		network::simulated::core::tests::symmetric \
-		network::simulated::core::tests::extreme_rate
+	    network::simulated::core::tests::symmetric \
+		network::simulated::token_bucket::tests::extreme_rate \
+		max_crashes \
+		three_nodes_crash
 }
 
 if [ $# -gt 0 ] && [ $1 == "slow" ]; then


src/consensus/block_producer.rs
```diff
@@ -292,41 +311,50 @@ where
...
+                    duration_left.saturating_sub(duration)
+                } else {
+                    duration_left.saturating_sub(self.delta_first_slice)
+                };
+                if left > Duration::ZERO {
+                    maybe_duration = Some(left);
+                } else {
+                    maybe_duration = None;
+                }
+            }
```
```suggestion
            let (payload, maybe_duration) =
                produce_slice_payload(&self.txs_receiver, parent, time_for_slice).await;

            let maybe_duration = if slice_index.is_first() {
                let elapsed = maybe_duration.unwrap_or(self.delta_first_slice);
                duration_left.checked_sub(elapsed)
            } else {
                maybe_duration
            };

```

akhi3030: is this more readable?

qkniep: yes, much more readable, `checked_sub` should indeed be fine because `DELTA_FIRST_SLICE <= DELTA_BLOCK`, right? but there is still this somewhat ugly edge case that this could now return `Some(Duration::ZERO)`, especially during testing when both deltas are set to 0

akhi3030: ah darn... lol yes, didn't think of that.  Will probably never happen (TM) in production but will probably happen in testing.

akhi3030: > should indeed be fine because DELTA_FIRST_SLICE <= DELTA_BLOCK, right?

We should have `const_assert` for that if we don't already.

qkniep: we do, but I can also add an `assert` to `BlockProducer`

qkniep:
@akhi3030: So, I changed the part that you also commented about, in the hope of making it easier to read. You can have a look if I succeeded. Also, I had an idea: Maybe it might actually be easier to do this, if instead of returning `Option<Duration>` we actually just use `Duration` with `None` being replaced with `Duration::ZERO` because that's exactly the behavior we want, right?

src/consensus/block_producer.rs
```diff
@@ -373,7 +399,7 @@ where
         .unwrap()
         .len();
 
-    // Super hacky!!!  As long as the size of the txs vec fits in a single byte,
+    // HACK: Super hacky!!! As long as the size of the txs vec fits in a single byte,
```
akhi3030: nit.


```suggestion
    // HACK: As long as the size of the txs vec fits in a single byte,
```

src/consensus/block_producer.rs
```diff
@@ -281,52 +303,55 @@ where
...
         let mut duration_left = self.delta_block;
         for slice_index in SliceIndex::all() {
-            let parent = if slice_index.is_first() {
-                Some(parent_block_id)
-            } else {
-                None
-            };
-            let time_for_slice = if slice_index.is_first() {
+            let (payload, maybe_duration) = if slice_index.is_first() {
                 // make sure first slice is produced on time
```
```suggestion
                // make sure first slice is produced quickly enough so that other nodes do not generate the [`TimeoutCrashedLeader`] event
```

src/consensus/block_producer.rs
```diff
@@ -281,52 +303,55 @@ where
...
+                    Some(parent_block_id),
+                    time_for_slice,
+                )
+                .await;
+                let elapsed = self.delta_first_slice - maybe_duration.unwrap_or(Duration::ZERO);
+                let left = if elapsed >= duration_left {
+                    None
+                } else {
+                    Some(duration_left - elapsed)
+                };
```
```suggestion
                let elapsed = self.delta_first_slice - maybe_duration.unwrap_or(Duration::ZERO);
                let left = (elapsed < duration_left).then_some(duration_left - elapsed);

```

Is this easier to read?  I am not sure it is necessarily simpler to read.

src/consensus/block_producer.rs
```diff
@@ -281,52 +303,55 @@ where
...
-            let parent = if slice_index.is_first() {
-                Some(parent_block_id)
-            } else {
-                None
-            };
-            let time_for_slice = if slice_index.is_first() {
+            let (payload, maybe_duration) = if slice_index.is_first() {
                 // make sure first slice is produced on time
-                duration_left.min(self.delta_first_slice)
+                let time_for_slice = duration_left.min(self.delta_first_slice);
```
```suggestion
                let time_for_slice = self.delta_first_slice;
```

Based on the assertion above, this should always be the case?  And then the logic below is clearer as we will know that `maybe_duration` can be at most delta_first_slice.

src/consensus/block_producer.rs
qkniep: I actually like it

src/consensus/block_producer.rs
```diff
@@ -253,24 +272,23 @@ where
...
 
-            match self
-                .shred_and_disseminate(header, payload, maybe_duration)
-                .await?
-            {
-                Either::Left(block_hash) => return Ok((slot, block_hash)),
-                Either::Right(duration) => duration_left = duration,
+            match self.shred_and_disseminate(header, payload).await? {
+                Some(block_hash) => return Ok((slot, block_hash)),
+                None => duration_left = new_duration_left,
```
akhi3030: nit: add assert that `new_duration_left` is not 0.

src/consensus/block_producer.rs
```diff
@@ -281,52 +299,52 @@ where
...
```
akhi3030: add an assert like above.

src/consensus/block_producer.rs
```diff
@@ -405,8 +424,7 @@ where
             }
         }
         if slice_capacity_left < MAX_TRANSACTION_SIZE {
-            let duration_left = duration_left.saturating_sub(Instant::now() - start_time);
-            break Some(duration_left);
+            break duration_left.saturating_sub(Instant::now() - start_time);
```
```suggestion
            break duration_left.saturating_sub(start_time.elapsed());
```

akhi3030 (approved):
LGTM.  Minor comments so disabled auto-merge.

PR #131, opened 2025-08-25
tooling: turn `lint.sh` script into general `check.sh` to locally check if CI will pass

qkniep:


check.sh
```diff
@@ -0,0 +1,3 @@
+cargo clippy --all-targets --all-features -- -Dwarnings
```
akhi3030, line 1: clippy parameters have changed.  Is that expected?

lint.sh
```diff
@@ -1,3 +0,0 @@
-cargo clippy --all-targets --all-features -- -W clippy::pedantic -W clippy::nursery -W clippy::cargo -A clippy::module_name_repetitions -A clippy::missing_errors_doc
-cargo +nightly udeps
-tokei
```
akhi3030, line 3: Is this still being run somewhere?

akhi3030 (approved):

check.sh
qkniep: yes, they now match CI, before they were way stricter in `lint.sh`

lint.sh
qkniep: `udeps` has been replaced by `machete` and `tokei` just counts lines of codes and is no longer part of script

PR #134, opened 2025-08-26
fix: `ParentReadyTracker` and `FinalityTracker` genesis handling, cleanup

### LINKED ISSUE #133: ParentReady duplicate
Some of the liveness tests (e.g. `three_nodes_crash`) crashed by violating this assertion `assertion failed: !ready_ids.contains(&id)`. It basically means that the `ParentReadyTracker` is marking the same parent ready for the same slot more than once.

qkniep:
Closes #133.

### DIFF
--- a/src/consensus/pool/finality_tracker.rs
+++ b/src/consensus/pool/finality_tracker.rs
@@ -25,7 +25,6 @@ use crate::crypto::Hash;
 use crate::types::Slot;
 
 /// Tracks finality of blocks.
-#[derive(Default)]
 pub struct FinalityTracker {
     /// Current finalization status for each slot.
     status: BTreeMap<Slot, FinalizationStatus>,
@@ -274,6 +273,24 @@ impl FinalityTracker {
         // recurse through ancestors
         if let Some(parent) = self.parents.get(&implicitly_finalized) {
             self.handle_implicitly_finalized(implicitly_finalized.0, *parent, event);
+        }
+    }
+}
+
+impl Default for FinalityTracker {
+    /// Creates a new empty tracker.
+    ///
+    /// Initially, only the genesis block is considered (directly) finalized.
+    fn default() -> Self {
+        let mut status = BTreeMap::new();
+        status.insert(
+            Slot::genesis(),
+            FinalizationStatus::Notarized(Hash::default()),
+        );
+        Self {
+            status,
+            parents: BTreeMap::new(),
+            highest_finalized_slot: Slot::genesis(),
         }
     }
 }
--- a/src/consensus/pool/parent_ready_tracker.rs
+++ b/src/consensus/pool/parent_ready_tracker.rs
@@ -36,16 +36,16 @@ impl ParentReadyTracker {
     ///
     /// Returns a list of any newly connected parents.
     /// All of these will have the given block ID as the parent.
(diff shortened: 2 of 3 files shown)


src/consensus/pool/finality_tracker.rs
```diff
@@ -278,6 +277,24 @@ impl FinalityTracker {
...
 
+impl Default for FinalityTracker {
+    /// Creates a new empty tracker.
+    ///
+    /// Initially, only the genesis block is considered (directly) finalized.
+    fn default() -> Self {
+        let mut status = BTreeMap::new();
+        status.insert(
+            Slot::genesis(),
+            FinalizationStatus::Notarized(Hash::default()),
```
akhi3030, line 288: I wonder if at some point we should define a genesis block hash.  So that we are not just setting it to default at different places and maybe we got it wrong somewhere.

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -56,39 +56,27 @@ impl ParentReadyTracker {
...
-            }
-        }
-
         // find possible parents for future windows
         let mut potential_parents = SmallVec::<[BlockId; 1]>::new();
-
-        for s in slot.slots_in_window().filter(|s| *s <= slot).rev() {
-            let state = self.slot_state(s);
-            if s < slot {
+        let window_slots = marked_slot.slots_in_window();
```
akhi3030, line 74: github will not let me leave a sensible suggestion for this block.  Is the following code equivalent and cleaner?

```rust
        let mut potential_parents = SmallVec::<[BlockId; 1]>::from_slice(state.ready_block_ids());
        let window_slots = marked_slot.slots_in_window();
        for slot in window_slots.filter(|s| s < &marked_slot).rev() {
            let state = self.slot_state(slot);
            for nf in &state.notar_fallbacks {
                potential_parents.push((slot, *nf));
            }
            if !state.skip {
                break;
            }
            potential_parents.extend_from_slice(state.ready_block_ids());
        }
```

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -56,39 +56,27 @@ impl ParentReadyTracker {
...
+        for slot in window_slots.filter(|s| *s <= marked_slot).rev() {
```
akhi3030, line 75: I think it will be useful to add a comment about why `rev` is used.

src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -99,14 +87,17 @@ impl ParentReadyTracker {
...
+            let state = self.slot_state(slot);
+            if slot.is_start_of_window() {
+                for parent in &potential_parents {
+                    state.add_to_ready(*parent);
+                    newly_certified.push((slot, *parent));
+                }
             }
-            for parent in &potential_parents {
-                newly_certified.push((first_slot, *parent));
+            if !state.skip {
```
akhi3030, line 99: can we have a comment for why this check is happening for every slot but the check above is just for start of window.

akhi3030 (approved):
LGTM.  Minor comments.  Disabled auto-merge.

src/consensus/pool/parent_ready_tracker.rs
qkniep: it should be equivalent, but mh, not sure this is easier to understand

PR #140, opened 2025-08-26
refactor: make block producer more rational

### LINKED ISSUE #128: Handle optimistic handover more aligned with rational block producer
We might be able to do without this. Any rational block producer would want to prevent that from ever happening, instead waiting longer to produce their slices. Also, we might want to have a higher slice count limit for optimistic block production anyways. One generally reasonable strategy would be to wait with producing the last slice until the timeout runs out (so way after ParentReady is emitted). I will create a new issue for this.

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/pull/123#discussion_r2297945133_

qkniep:
Closes #128.

Before, the block producer handled the case where all slices are already produced and we haven't seen `ParentReady` yet. This is never rational behavior for the block producer. To this end, this is a refactor of the production of slices and blocks, such that it is more aligned with rational behavior and can be extended with arbitrary transaction packing strategies. Specifically, now slice production takes a preempt channel and a minimum duration, which can be used for pacing. The default block producer now waits on slice number `MAX_SLICES_PER_BLOCK / 2` until being explicitly preempted due to the `ParentReady` event.


src/consensus/block_producer.rs
```diff
@@ -184,48 +184,87 @@ where
...
-            } else {
-                None
-            };
+        let block_hash = self
+            .produce_slices(
+                slot,
+                parent,
+                // only start the DELTA_BLOCK timer later, when ParentReady event is seen
+                Duration::MAX,
+                Some(&mut parent_ready_receiver),
```
akhi3030: If the receiver is some, then the duration is always max and if the receiver is none then the duration is always self.delta_block.  Maybe we just pass the receiver and let `produce_slices` figure out what the duration should be.

src/consensus/block_producer.rs
```diff
@@ -184,48 +184,87 @@ where
...
                 // TODO: this can be made more accurate, only needed if production of first slice
                 // still takes more than delta_first_slice after we saw ParentReady, not if:
                 // 1. first slice is produced before ParentReady is seen, OR
                 // 2. first slice finishes at most delta_first_slice after ParentReady is seen
-                duration_left.min(self.delta_first_slice)
+                self.delta_first_slice
             } else {
                 // cap timeout for each slice to `DELTA_BLOCK`
                 // makes sure optimistic block production yields before timeout would expire
                 duration_left.min(self.delta_block)
```
akhi3030: Logically, duration_left should always be <= delta_block.  I would replace this with assertions (or debug assertions).

src/consensus/block_producer.rs
```diff
@@ -184,48 +184,87 @@ where
...
+            parent_slot,
+        );
+
+        let block_hash = self
+            .produce_slices(slot, parent, self.delta_block, None)
+            .await?;
+        Ok((slot, block_hash))
+    }
+
+    async fn produce_slices(
```
akhi3030, line 223: If we are introducing this function, we can probably get rid of the two produce_block_* functions above.

I am still on the fence which version I like better.  I think having 2 produce_block_* is easier to read but there is more code duplication.  If you find that combining works better for you, then we can do it like that.  If we do add more special casing here, we should consider if switching to separate functions is more readable or not.

src/consensus/block_producer.rs
```diff
@@ -184,48 +184,87 @@ where
...
+        &self,
+        slot: Slot,
+        parent: BlockId,
+        mut duration_left: Duration,
+        mut parent_ready_receiver: Option<&mut oneshot::Receiver<BlockId>>,
+    ) -> Result<Hash> {
+        for slice_index in SliceIndex::all() {
+            let (parent_slot, parent_hash) = parent;
+            let slice_parent = slice_index.is_first().then_some(parent);
```
akhi3030: I would move these bindings closer to where they are actually needed.  i.e.:

- the first one to right before the comparison with the new_hash.
- the second one to right before calling produce_slice_payload.

src/consensus/block_producer.rs
```diff
@@ -184,48 +184,87 @@ where
...
                     res = &mut produce_slice_future => {
-                        let (payload, _new_duration_left) = res;
+                        let (payload, _time_elapsed) = res;
                         // ParentReady event still not seen, do not start DELTA_BLOCK timer yet
                         (payload, Duration::MAX)
                     }
-                    res = &mut parent_ready_receiver => {
+                    res = parent_ready_receiver.as_mut().unwrap() => {
                         // Got ParentReady event while producing slice.
                         // It's a NOP if we have been using the same parent as before.
```
akhi3030: if we set parent_ready_receiver to None here, we should be able to get rid of the two unwraps above.

src/consensus/block_producer.rs
```diff
@@ -240,96 +279,30 @@ where
...
-            match self.shred_and_disseminate(header, payload).await? {
-                Some(block_hash) => return Ok((slot, block_hash)),
-                None => {
-                    assert!(!new_duration_left.is_zero());
-                    duration_left = new_duration_left;
-                }
+            if let Some(ref parent_rcv) = parent_ready_receiver
+                && is_last
+            {
+                assert!(parent_rcv.is_terminated());
```
akhi3030, line 320: I am not sure if I follow the logic that this condition should always hold.  slice_index.is_max() can be true without us seeing the parent ready event, no?

src/consensus/block_producer.rs
```diff
@@ -412,7 +389,7 @@ where
         let sleep_duration = duration_left.saturating_sub(Instant::now() - start_time);
```
```suggestion
        let sleep_duration = duration_left.saturating_sub(start_time.elapsed());
```

src/consensus/block_producer.rs
```diff
@@ -412,7 +389,7 @@ where
...
         let res = tokio::select! {
             () = tokio::time::sleep(sleep_duration) => {
-                break Duration::ZERO;
+                break duration_left;
```
akhi3030: I don't think I understand this change.  Didn't we use up all the time in producing the slice here?  So shouldn't we be returning 0?

akhi3030 (commented):
I would say the following comments are the most essential ones:

- https://github.com/qkniep/alpenglow/pull/140/files#r2303478194
- https://github.com/qkniep/alpenglow/pull/140/files#r2303486923

We should align on those before we worry about the other comments which are mainly cosmetic.

Also, I do not see any additional logic to handle #128.  Is this just a commit to refactor stuff and actual issue would be handled in a different PR?

src/consensus/block_producer.rs
qkniep: yes, one of the end goals of this PR is that this will always hold, that's not yet the case, because for that to be the case we need a way of pacing slice production

qkniep:
I was thinking mostly in general if this structure makes more or less sense: To have minimal logic in the two `produce_block_XXX()` functions and have this separate `produce_slices()` function. Now I also changed `produce_slice()` to no longer wait for a specific time but to instead be able to to preempt it from the outside via a `oneshot::channel()`. This should allow us to have more control over how long to wait on each slice. However, actually changing that and thus making that one assertion correct is still open at the moment.

src/consensus/block_producer.rs
qkniep: not in the parent-not-ready case, where it is `Duration::MAX` initially

src/consensus/block_producer.rs
qkniep: this function changed from returning time left to returning time elapsed

src/consensus/block_producer.rs
```diff
@@ -144,18 +145,22 @@ where
                 }
             };
             debug!(
-                "produced block {} in {} ms",
+                "produced block {} ({} txs) in {} ms",
```
```suggestion
                "produced first block in window {} ({} txs) in {} ms",
```

Just to differentiate it a bit from the other log statement.

src/consensus/block_producer.rs
```diff
@@ -122,7 +123,7 @@ where
 
             // produce first block
             let start = Instant::now();
-            let mut block_id = match slot_ready {
+            let (mut block_id, num_txs) = match slot_ready {
```
akhi3030, line 126: Are there plans to use `num_txs` for something in the future?  Currently, we just log it.  I am wondering if that is terribly useful at this point in time.

src/consensus/block_producer.rs
```diff
@@ -170,11 +175,11 @@ where
     pub(super) async fn produce_block_parent_not_ready(
```
akhi3030, line 177: not sure if you saw my other comment on an older version of the PR.  I am wondering if we can get rid of this function and `produce_block_parent_ready`.  They are not doing much.

src/consensus/block_producer.rs
```diff
@@ -184,103 +189,10 @@ where
...
-                Some(block_hash) => return Ok((slot, block_hash)),
-                None => {
-                    assert!(!new_duration_left.is_zero());
-                    duration_left = new_duration_left;
-                }
-            }
-        }
-        unreachable!()
+        let (block_hash, num_txs) = self
+            .produce_slices(slot, parent, Some(&mut parent_ready_receiver))
```
akhi3030, line 195: if we do keep this function around, I wonder if we get get `produce_slices` to return `(BlockId, usize)` and not just the hash.  Then we don't need the extra line here.

src/consensus/block_producer.rs
```diff
@@ -170,11 +175,11 @@ where
     pub(super) async fn produce_block_parent_not_ready(
         &self,
         slot: Slot,
-        parent_block_id: BlockId,
+        parent: BlockId,
         mut parent_ready_receiver: oneshot::Receiver<BlockId>,
-    ) -> Result<BlockId> {
+    ) -> Result<(BlockId, usize)> {
```
akhi3030, line 182: we should add a comment about what `usize` is.  The type itself doesn't tell us very much.

src/consensus/block_producer.rs
```diff
@@ -289,47 +201,130 @@ where
...
+        // only start the DELTA_BLOCK timer when ParentReady event is seen
+        let mut duration_left = match parent_ready_receiver {
+            None => self.delta_block,
+            Some(_) => Duration::MAX,
+        };
+        let mut txs_in_block = 0;
+
+        for slice_index in SliceIndex::all() {
+            let slice_parent = slice_index.is_first().then_some(parent);
+            // HACK: use of `MAX_SLICES_PER_BLOCK / 2`, maybe introduce `MAX_SLICES_FIRST_SLOT`
```
akhi3030, line 238: Either here or in the documentation of `produce_slice_payload` we should explain the intent of min_time_for_slice.

src/consensus/block_producer.rs
```diff
@@ -289,47 +201,130 @@ where
...
+                        }
+                        res = parent_rcv => {
+                            parent_ready_receiver = None;
+                            // Got ParentReady event while producing slice.
+                            // It's a NOP if we have been using the same parent as before.
+                            let start = Instant::now();
+                            let (new_slot, new_hash) = res.expect("sender dropped");
+                            let (payload, _time_elapsed, txs_in_slice) = tokio::select! {
+                                () = tokio::time::sleep(self.delta_first_slice) => {
+                                    preempt_sender.send(()).unwrap();
```
akhi3030, line 274: Hmm... I think this `unwrap` is not safe.

`produce_slice_payload` might return without receiving a value on the channel i.e. dropping the channel.  And if this `send` were to happen after the drop, it will panic.

src/consensus/block_producer.rs
```diff
@@ -430,15 +428,31 @@ where
...
-    // TODO: not accounting for this potentially expensive operation in duration_left calculation above.
+    let num_txs = txs.len();
     let txs = bincode::serde::encode_to_vec(&txs, bincode::config::standard())
         .expect("serialization should not panic");
     let payload = SlicePayload::new(parent, txs);
-    (payload, ret)
+
+    // idle if necessary
+    // TODO: instead actually spend more time waiting for and selecting transactions
+    if start_time.elapsed() < min_duration && !preempt_receiver.is_terminated() {
```
akhi3030, line 439: This would be true if the slice is already full?  In that case, if we receive more txs, where will be put them?  

If a potential idea is to buffer up the additional txs in an internal queue, we will have to be careful not to buffer up more than we will be able to fit in the remaining slices of this slot and potentially remaining slots of the leader window.

src/consensus/block_producer.rs
```diff
@@ -372,22 +370,24 @@ where
...
-// TODO: extend docstring
-/// Returns
+/// Produces a slice (payload only).
+///
+/// Packs transactions received from the network into the slice until either:
+/// - The slice is full (accounting for encoding of parent), OR
+/// - there is an incoming message on the `preempt_receiver`.
+///
+/// Returns the slice payload, amount of time elapsed, and number of transactions packed.
 async fn produce_slice_payload<T>(
```
akhi3030, line 388: I am wondering if having two versions of this function would make them simpler albeit with some code duplication.

Parent ready version:
`produce_slice_payload_parent_ready(txs_receiver, parent:Option<BlockId>, duration) -> (SlicePayload, Duration, usize)`: it doesn't need a preemption channel or a minimum duration.  It needs a maximum duration.  

Parent not ready version: would probably not change very much but it we will have a simpler to understand more straightforward one above.

akhi3030 (commented):
Some initial feedback.  Happy to review again once we have alignment on how to move forward.

qkniep:
Closing in favor of #381.

PR #143, opened 2025-08-27
wip: improve caching of build artifacts in CI

akhi3030:
Comparing https://github.com/qkniep/alpenglow/actions/runs/17241093422/job/48918238882 a run before this PR and https://github.com/qkniep/alpenglow/actions/runs/17262669277/job/48987527319?pr=143 a run on this PR, we are caching the crates but we are not caching the build artifacts.  Needs more investigation.


akhi3030:
@qkniep: my current theory is the following:

- We are actually caching and restoring the dependencies and the target directory
- When CI runs, it downloads and untars the cache
- then we run `cargo *` commands which inspects the file modification timestamps to determine what to recompile
- because how the cache was untared, it ends up recompiling things still
- I experimented with `CARGO_INCREMENTAL` as well (the current state of the PR) but that doesn't seem to have made a difference

We could explore [sccache](https://github.com/mozilla/sccache?tab=readme-ov-file) some people have reported getting some benefits with that.

PR #150, opened 2025-08-28
feat: faster Merkle tree construction, more max nodes, handle edge cases

### LINKED ISSUE #148: Use `EMPTY_ROOTS` in `MerkleTree::new()`
The pre-calculated roots of empty sub-trees in `EMPTY_ROOTS` can be used to speed up creation of the Merkle tree by up to a factor of 2.

This has limited impact as long as we choose `TOTAL_SHREDS` as a power of two, because the most Merkle tree calculations we do are for the Merkle root of each slice.

qkniep:
Handles some open tasks in the `crypto::merkle` module. Some out-of-bounds accesses have assertions now. Also increased max number of leaves from 2^24 to 2^32, providing the efficient function used to calculate those.

Maybe most important, we now use the pre-calculated `EMPTY_ROOTS` to speed up the constructor (up to a factor of two if the second half of the tree is almost empty).

Closes #148.

### DIFF
--- a/src/crypto/merkle.rs
+++ b/src/crypto/merkle.rs
@@ -217,6 +217,8 @@ fn hash_pair(left: Hash, right: Hash) -> Hash {
 
 #[cfg(test)]
 mod tests {
+    use rand::prelude::*;
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/crypto/merkle.rs
```diff
@@ -63,55 +83,78 @@ const EMPTY_ROOTS: [Hash; 24] = [
...
             let leaf_hash = hash_leaf(leaf);
             nodes.push(leaf_hash);
         }
-
-        // PERF: use `EMPTY_ROOTS` to get rid of these explicit empty leaves
-        let empty_leaf_hash = hash_leaf(&[]);
-        while !nodes.len().is_power_of_two() {
-            nodes.push(empty_leaf_hash);
-        }
+        levels.push((0, nodes.len() as u32));
```
akhi3030, line 128: I highly doubt it'll ever happen but if there is an overflow, we will get weird behaviour.  


```suggestion
        levels.push((0, nodes.len().try_into().unwrap()));
```

akhi3030 (approved):

PR #156, opened 2025-08-29
wip: use `IpAddr` instead of strings for addresses

akhi3030:
incomplete.


akhi3030:
this of course doesn't make any sense.  We need the port number too so we need to use `SocketAddr`.

PR #157, opened 2025-08-29
refactor: use `SocketAddr` instead of `String` for storing addresses

akhi3030:
This makes things a bit more ergonomic; reduces string parsing when sending messages; etc.

### DIFF
--- a/src/consensus/cert.rs
+++ b/src/consensus/cert.rs
@@ -560,11 +560,13 @@ mod tests {
     use super::*;
     use crate::crypto::aggsig::SecretKey;
     use crate::crypto::signature;
+    use crate::network::dontcare_sockaddr;
 
     fn create_signers(signers: u64) -> (Vec<SecretKey>, Vec<ValidatorInfo>) {
         let mut sks = Vec::new();
         let mut voting_sks = Vec::new();
         let mut info = Vec::new();
+        let sockaddr = dontcare_sockaddr();
 
         for i in 0..signers {
             sks.push(signature::SecretKey::new(&mut rand::rng()));
@@ -574,9 +576,9 @@ mod tests {
                 stake: 1,
                 pubkey: sks.last().unwrap().to_pk(),
                 voting_pubkey: voting_sks.last().unwrap().to_pk(),
-                all2all_address: String::new(),
-                disseminator_address: String::new(),
-                repair_address: String::new(),
+                all2all_address: sockaddr,
+                disseminator_address: sockaddr,
+                repair_address: sockaddr,
             });
         }
 
--- a/src/consensus/cert.rs
+++ b/src/consensus/cert.rs
@@ -566,7 +566,6 @@ mod tests {
         let mut sks = Vec::new();
         let mut voting_sks = Vec::new();
         let mut info = Vec::new();
-        let sockaddr = dontcare_sockaddr();
 
         for i in 0..signers {
             sks.push(signature::SecretKey::new(&mut rand::rng()));
@@ -576,9 +575,9 @@ mod tests {
                 stake: 1,
                 pubkey: sks.last().unwrap().to_pk(),
                 voting_pubkey: voting_sks.last().unwrap().to_pk(),
-                all2all_address: sockaddr,
-                disseminator_address: sockaddr,
-                repair_address: sockaddr,
+                all2all_address: dontcare_sockaddr(),
+                disseminator_address: dontcare_sockaddr(),
+                repair_address: dontcare_sockaddr(),
             });
         }
 
--- a/src/disseminator/rotor/sampling_strategy.rs
+++ b/src/disseminator/rotor/sampling_strategy.rs
@@ -656,6 +656,7 @@ mod tests {
     use crate::crypto::aggsig;
     use crate::crypto::signature::SecretKey;
     use crate::disseminator::turbine::WeightedShuffle;
+    use crate::network::dontcare_sockaddr;
     use crate::network::simulated::stake_distribution::VALIDATOR_DATA;
     use crate::shredder::TOTAL_SHREDS;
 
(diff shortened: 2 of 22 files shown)


src/consensus/cert.rs
```diff
@@ -574,9 +576,9 @@ mod tests {
                 stake: 1,
                 pubkey: sks.last().unwrap().to_pk(),
                 voting_pubkey: voting_sks.last().unwrap().to_pk(),
-                all2all_address: String::new(),
-                disseminator_address: String::new(),
-                repair_address: String::new(),
+                all2all_address: sockaddr,
```
qkniep: maybe call `dontcare_sockaddr()` here directly, instead of using `sockaddr` variable

src/disseminator/rotor/sampling_strategy.rs
```diff
@@ -664,14 +665,15 @@ mod tests {
...
+            let sockaddr = dontcare_sockaddr();
             validators.push(ValidatorInfo {
                 id: i,
                 stake: 1,
                 pubkey: sk.to_pk(),
                 voting_pubkey: voting_sk.to_pk(),
-                all2all_address: String::new(),
-                disseminator_address: String::new(),
-                repair_address: String::new(),
+                all2all_address: sockaddr,
```
qkniep: same

src/disseminator/turbine.rs
```diff
@@ -244,14 +244,15 @@ mod tests {
...
+            let sock = dontcare_sockaddr();
             validators.push(ValidatorInfo {
                 id: i,
                 stake: 1,
                 pubkey: sks[i as usize].to_pk(),
                 voting_pubkey: voting_sks[i as usize].to_pk(),
-                all2all_address: String::new(),
-                disseminator_address: String::new(),
-                repair_address: String::new(),
+                all2all_address: sock,
```
qkniep: same

src/network/simulated/stake_distribution.rs
```diff
@@ -202,9 +204,9 @@ pub fn validators_from_validator_data(
                 stake,
                 pubkey: sk.to_pk(),
                 voting_pubkey: voting_sk.to_pk(),
-                all2all_address: String::new(),
-                disseminator_address: String::new(),
-                repair_address: String::new(),
+                all2all_address: sockaddr,
```
qkniep: same

src/network/simulated/stake_distribution.rs
```diff
@@ -231,9 +233,9 @@ pub fn validators_from_validator_data(
...
```
qkniep: same

src/network/udp.rs
```diff
@@ -89,22 +78,24 @@ impl Network for UdpNetwork {
...
+    use std::net::IpAddr;
+
     use super::*;
 
     #[tokio::test]
     async fn ping() {
         let socket1 = UdpNetwork::new_with_any_port();
         let socket2 = UdpNetwork::new_with_any_port();
-        let addr1 = format!("127.0.0.1:{}", socket1.port());
+        let addr1 = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), socket1.port());
```
qkniep: could replace with `localhost_ip_sockaddr()` call

src/network/udp.rs
```diff
@@ -113,13 +104,13 @@ mod tests {
     async fn ping_pong() {
         let socket1 = UdpNetwork::new_with_any_port();
         let socket2 = UdpNetwork::new_with_any_port();
-        let addr1 = format!("127.0.0.1:{}", socket1.port());
-        let addr2 = format!("127.0.0.1:{}", socket2.port());
+        let addr1 = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), socket1.port());
```
qkniep: same

src/main.rs
```diff
@@ -94,9 +96,9 @@ fn create_test_nodes(count: u64) -> Vec<TestNode> {
             stake: 1,
             pubkey: sks[id as usize].to_pk(),
             voting_pubkey: voting_sks[id as usize].to_pk(),
-            all2all_address: format!("127.0.0.1:{port}"),
-            disseminator_address: format!("127.0.0.1:{}", port + 1),
-            repair_address: format!("127.0.0.1:{}", port + 2),
+            all2all_address: SocketAddr::new(ipaddr, port),
```
qkniep: here as well. call `localhost_ip_sockaddr()`?

src/network.rs
```diff
@@ -156,32 +156,23 @@ pub enum NetworkError {
...
-            .parse()
-            .map_err(|_| NetworkError::MalformedAddress)
-    }
+/// Returns a [`SocketAddr`] bound to the localhost IPv4 and given port.
+pub fn localhost_ip_sockaddr(port: u16) -> SocketAddr {
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), port)
+}
+
+/// Returns a [`SocketAddr`] that could be bound any arbitrary IP and port.
+pub fn dontcare_sockaddr() -> SocketAddr {
```
qkniep, line 181: this should only be used in testing, right? so maybe use `#[cfg(test)]` or at least write a note in the docstring

src/network.rs
```diff
@@ -156,32 +156,23 @@ pub enum NetworkError {
...
-    fn parse_addr(str: impl AsRef<str>) -> Result<Self::Address, NetworkError>
-    where
-        Self::Address: FromStr,
-    {
-        str.as_ref()
-            .parse()
-            .map_err(|_| NetworkError::MalformedAddress)
-    }
+/// Returns a [`SocketAddr`] bound to the localhost IPv4 and given port.
+pub fn localhost_ip_sockaddr(port: u16) -> SocketAddr {
```
qkniep, line 173: add a comment about port 0, or have an extra function like `localhost_ip_sockaddr_any_port()`

qkniep (approved):

src/network.rs
qkniep: ok, maybe you can ignore this, we currently only use port 0 with this function for simulated network, right?

akhi3030: Yes, we are only using simulated network for now.  I tried to change it so that calling this function with port=0 panics.  Then started updating the various test code but some tests were hanging.  I think I will add a todo here to fix for later and merge for now.

PR #158, opened 2025-08-29
fix: finalization slot/block, naming around certs/finalization, adapt tests

qkniep:
Previously, the basically meaningless notion of a finalized slot without a block was wrongfully exposed and used (i.e. the case when we have a finalization cert but don't yet have the notarization cert for a block). This is now fixed, only exposing the notion of notarized block as having fast finalization or final+notar certs outside of `FinalityTracker`. To go along with this, some of the naming around finalization and certificates has been adapted to be more accurate and less error prone.

### DIFF
--- a/src/consensus/pool/finality_tracker.rs
+++ b/src/consensus/pool/finality_tracker.rs
@@ -31,6 +31,9 @@ pub struct FinalityTracker {
     /// Maps blocks to their parents.
     parents: BTreeMap<BlockId, BlockId>,
     /// The highest finalized slot so far.
+    ///
+    /// This means that slot has a fast finalization OR finalization + notarization.
+    /// Also, all prior slots are finalized (directly or implicitly) OR implicitly skipped.
     highest_finalized_slot: Slot,
 }
 
@@ -39,7 +42,7 @@ pub struct FinalityTracker {
 pub enum FinalizationStatus {
     /// Block with given hash is notarized, but slot is not yet (known to be) finalized.
     Notarized(Hash),
-    /// Slot is finalized, but notarized block is not yet known.
+    /// Slot is known to be finalized, but we are missing the notarization certificate.
     FinalPendingNotar,
     /// Slot is finalized, and notarized block is known to have the given hash.
     Finalized(Hash),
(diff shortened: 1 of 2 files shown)


src/consensus/pool/finality_tracker.rs
```diff
@@ -40,9 +40,9 @@ pub enum FinalizationStatus {
     /// Block with given hash is notarized, but slot is not yet (known to be) finalized.
```
akhi3030, line 43: cannot leave a comment further above.  Is the meaning of FinalityTracker::highest_finalized_slot also changed now?  Maybe in comments or otherwise, should we clarify precisely what it means?

src/consensus/pool/finality_tracker.rs
```diff
@@ -166,31 +165,32 @@ impl FinalityTracker {
...
-                    self.handle_finalized_block((slot, block_hash), &mut event);
-                    self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
-                    event
-                }
-                FinalizationStatus::ImplicitlySkipped => unreachable!("consensus safety violation"),
+        let old = self
+            .status
+            .insert(slot, FinalizationStatus::FinalPendingNotar);
+        let Some(status) = old else {
+            return FinalizationEvent::default();
```
akhi3030, line 175: previously we were updating `self.highest_finalized_slot`, we don't need to anymore?

src/consensus/pool/finality_tracker.rs
qkniep: yes, the idea is that `highest_finalized_slot` now refers only to a slot that has a fast finalization or final+notar certs... the only remaining notion of the old meaning is `FinalizationStatus::FinalPendingNotar`... I explain this in the comment for `highest_finalized_slot()` (the function not the variable), but I can also make it more clear for the variable

src/consensus/pool/finality_tracker.rs
qkniep: yes, now we should only update it once a block is actually finalized (not as soon as we have the finalization certificate, but not yet the notar), this is now done in `handle_finalized_block()`

akhi3030 (approved):

PR #160, opened 2025-08-30
wip

akhi3030:


src/bin/performance_test.rs
```diff
@@ -79,15 +79,17 @@ async fn create_test_nodes(count: u64) -> Vec<TestNode> {
         voting_sks.push(aggsig::SecretKey::new(&mut rng));
         let all2all_address = localhost_ip_sockaddr((3 * id).try_into().unwrap());
         let disseminator_address = localhost_ip_sockaddr((3 * id + 1).try_into().unwrap());
-        let repair_address = localhost_ip_sockaddr(udp_networks[id as usize].port());
+        let repair_request_address = localhost_ip_sockaddr(udp_networks[id as usize].port());
```
akhi3030, line 82: need to fix these.

src/all2all/robust.rs
```diff
@@ -94,7 +94,8 @@ mod tests {
                 voting_pubkey: voting_sk.to_pk(),
                 all2all_address: localhost_ip_sockaddr(i.try_into().unwrap()),
                 disseminator_address: dontcare_sockaddr(),
-                repair_address: dontcare_sockaddr(),
+                repair_request_address: dontcare_sockaddr(),
+                repair_response_address: dontcare_sockaddr(),
```
qkniep, line 98: These names could become kind of confusing. Because `repair_request_address` and `repair_request_network` will be used to receive `RepairRequest`s and send `RepairResponse`s, right? Not sure there is a better naming for these, but maybe at least a comment somewhere (e.g. in `ValidatorInfo` and `Alpenglow`) making this very explicit could help.

akhi3030: @qkniep: what's interesting is that `repair_response_address` is used in precisely one place right now.  https://github.com/qkniep/alpenglow/pull/160/files#diff-123edf9b5a358c514cb89166397418d6b8e215f014a198bfb7ac30dd5e22b59bR176.  i.e. to reply to repair requests.

Conceptually, we can generalise things a bit further here.  When a node starts repair, they can pick a port number for receiving the responses.  This way, they could really optimise which repair they want to prioritise and spread the load over multiple ports if they are repairing a lot.

I don't think we need to do the above right now or any time soon.  But what we could consider doing now is that the node indicates in the repair request on which address they want to receive the response.  Doing that will allow us to remove `repair_response_address` from `ValidatorInfo` and potentially also allow us to remove it from `Alpenglow::new()` if we let `Repair` pick "any port" for receiving the responses.

Doing this would have the potential issue that a malicious node A sends a request to node B and asks the reply be sent to the victim C which gets bombarded by network packets.  However, I would not worry about this attack too much.  There are simpler and cheaper ways of bombarding a node with arbitrary UDP packets.  But if we are really worried about it, we could say that the repair replies always go back to the same IP address but the port number is specified in the request.

qkniep: Mh, I think this approach has more downsides than just the DoS vector that you outlined: (1) it makes symmetric authentication and blacklisting of nodes harder, (2) randomly starting to use new ports in UDP might be a no-go depending on NAT/firewall setup. It might still be viable, but it's not straightforward.

akhi3030: Agreed, we can keep the current approach then.  I'll think about better naming / comments.

akhi3030: Added some documentation.

src/consensus.rs
```diff
@@ -93,13 +93,18 @@ where
     T: Network + Sync + Send + 'static,
 {
     /// Creates a new Alpenglow consensus node.
+    ///
+    /// `repair_network` - Network from which the node sends [`RepairRequest`] messages and receives [`RepairResponse`] messages.
+    /// `repair_request_network` - Network where the node receives [`RepairRequest`] messages and sends [`RepairResponse`] messages.
     #[must_use]
+    #[allow(clippy::too_many_arguments)]
```
qkniep, line 100: ouch

akhi3030: Yeah we will get some of these on constructors.  Hopefully as long as it is just constructors, that is not a bad sign but if we start seeing them on other functions, then we should rethink things.

src/repair.rs
```diff
@@ -68,15 +68,126 @@ pub enum RepairResponse {
...
+    pub async fn run(&self) {
+        loop {
+            match self.network.receive().await.unwrap() {
+                NetworkMessage::Repair(message) => match message {
+                    RepairMessage::Request(request, sender) => {
+                        self.answer_request(request, sender).await.unwrap()
+                    }
+                    msg => panic!(
+                        "validator: {} Unexpected msg tpe: {:?}",
+                        self.epoch_info.own_id, msg
```
qkniep: I think, in general, we should not log validator ID manually. Instead, we might want to find a way to use something like spans to automatically tag the validator ID or split the log outputs. We might need to go back to `tracing` to do that...

src/repair.rs
```diff
@@ -68,15 +68,126 @@ pub enum RepairResponse {
...
+    /// Looks up the corresponding data in `self.blockstore` and sends replies.
+    pub async fn run(&self) {
+        loop {
+            match self.network.receive().await.unwrap() {
+                NetworkMessage::Repair(message) => match message {
+                    RepairMessage::Request(request, sender) => {
+                        self.answer_request(request, sender).await.unwrap()
+                    }
+                    msg => panic!(
+                        "validator: {} Unexpected msg tpe: {:?}",
```
qkniep:
```suggestion
                        "validator: {} Unexpected msg type: {:?}",
```

src/repair.rs
```diff
@@ -68,15 +68,126 @@ pub enum RepairResponse {
...
+                        self.epoch_info.own_id, msg
+                    );
+                }
+            }
+        }
+    }
+
+    /// Tries to answer the given repair request.
+    ///
+    /// If we do not have the block the request refers to, the request is ignored.
```
qkniep:
```suggestion
    /// If we do not have the necessary information in blockstore, the request is ignored.
```
e.g. for a `Shred` request having received the specific shred through repair is enough

src/repair.rs
```diff
@@ -327,7 +379,7 @@ impl<N: Network> Repair<N> {
     /// # Errors
     ///
     /// Returns [`std::io::Error`] if the underlying network fails.
-    async fn receive(&self) -> io::Result<RepairMessage> {
+    async fn receive(&self) -> io::Result<RepairResponse> {
```
qkniep, line 376: doc string needs to be updated

src/repair.rs
```diff
@@ -340,8 +392,17 @@ impl<N: Network> Repair<N> {
                 }
             };
             match msg {
-                NetworkMessage::Repair(r) => return Ok(r),
-                m => warn!("unexpected message type for repair: {m:?}"),
+                NetworkMessage::Repair(msg) => match msg {
+                    RepairMessage::Response(msg) => return Ok(msg),
+                    rest => panic!(
+                        "validator {}: unexpected msg type: {rest:?}",
```
qkniep: same as above: logging validator ID

src/repair.rs
```diff
@@ -498,18 +585,23 @@ mod tests {
...
-        let (repair_channel, blockstore, other_network, sk) = create_repair_instance().await;
+        let (repair_channel, blockstore, other_network_request, _other_network_reply, sk) =
+            create_repair_instance().await;
 
         // create a block to repair
         let slot = Slot::genesis().next();
         let (block_hash, merkle_tree, shreds) = create_random_shredded_block(slot, num_slices, &sk);
         let block_to_repair = (slot, block_hash);
 
+        println!("sending repair on channel");
```
qkniep: remove prints?

qkniep (approved):
LGTM, apart from above comments

PR #166, opened 2025-09-01
wip: ingest missing shreds when reconstructing slice

qkniep:


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -219,27 +219,39 @@ impl BlockData {
     ///
     /// Returns `true` if a slice was reconstructed, `false` otherwise.
     fn try_reconstruct_slice(&mut self, slice: SliceIndex) -> bool {
-        let slice_shreds = self.shreds.get(&slice).unwrap();
+        let slice_shreds = self.shreds.get_mut(&slice).unwrap();
         if self.slices.contains_key(&slice) {
             return false;
         }
```
```suggestion
if self.slices.contains_key(&slice) {
return false;
}
        let slice_shreds = self.shreds.get_mut(&slice).unwrap();
```

akhi3030: saves a lookup in case of early return.

src/shredder.rs
```diff
@@ -481,6 +508,46 @@ pub fn data_and_coding_to_output_shreds(
...
+    data: Vec<DataShred>,
+    coding: Vec<CodingShred>,
+    tree: MerkleTree,
+    leader_signature: Signature,
+) -> Vec<Shred> {
+    let mut shreds = Vec::with_capacity(data.len() + coding.len());
+    let num_data_shreds = data.len();
+    let merkle_root = tree.get_root();
+
+    for d in data.into_iter() {
```
akhi3030:
```suggestion
    for d in data {
```

src/shredder.rs
```diff
@@ -481,6 +508,46 @@ pub fn data_and_coding_to_output_shreds(
...
+        let merkle_path = tree.create_proof(d.0.index_in_slice);
+        shreds.push(Shred {
+            payload_type: ShredPayloadType::Data(d.0),
+            merkle_root,
+            merkle_root_sig: leader_signature,
+            merkle_path,
+        });
+    }
+    for mut c in coding.into_iter() {
```
akhi3030:
```suggestion
    for mut c in coding {
```

src/shredder.rs
```diff
@@ -247,7 +250,7 @@ impl Shredder for RegularShredder {
         Ok(data_and_coding_to_output_shreds(data, coding, sk))
     }
 
-    fn deshred(shreds: &[Shred]) -> Result<Slice, DeshredError> {
+    fn deshred(shreds: &[Shred]) -> Result<(Slice, Vec<Shred>), DeshredError> {
```
akhi3030, line 253: future work to consider: I wonder how cpu intense these calls are and if we should be spawn_blocking them.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -219,27 +219,39 @@ impl BlockData {
...
+        // insert reconstructed slice and shreds
         self.slices.insert(slice, reconstructed_slice);
+        for shred in reconstructed_shreds {
+            let exists = slice_shreds
+                .iter()
+                .any(|s| s.payload().index_in_slice == shred.payload().index_in_slice);
+            if !exists {
+                slice_shreds.push(shred);
+            }
+        }
```
```suggestion
        std::mem::swap(slice_shreds, &mut reconstructed_shreds);
```

Is this equivalent?

qkniep: for now yes, we could optimize this later, to only reconstruct shreds that are not yet in blockstore

akhi3030: Yes, indeed, that will be an important optimisation to figure out in the future.

PR #170, opened 2025-09-03
refactor: improvements to how the networks are initialized in `liveness::create_test_nodes()`

akhi3030:
Introduces a struct to help track which network is for which component and not having to do a lot of indexing into `Vec` or `pop`ing from a `VecDeque` in the right order.


akhi3030:
If this PR is accepted, in a subsequent PR, I can move the struct to a more common place so that the tests in `src/main.rs` can also benefit from it.

qkniep (approved):
LGTM

PR #171, opened 2025-09-03
refactor: combine the two definitions of `create_test_nodes` into one

akhi3030:
We had a similar definition in two places: `src/main.rs` and `tests/liveness.rs`.

Unfortunately, the shared code has to go into `src/lib.rs`.


qkniep (approved):
LGTM

PR #177, opened 2025-09-04
RFC: `Network::receive` only returns the desired type of message

akhi3030:
Currently, whenever we are doing `Network::receive()`, we need to handle the different variants that we do not care about.  In most cases, we ignore the other variants and we simply call `receive()` again.

The idea here is to introduce a way to get the `Network` to keep looping till it receives the variant we care about.  The PR is still incomplete.  I wanted to get some feedback on whether this idea makes sense or not before I do the work to finish it up.

We introduce a new trait `FromNetworkMessage`.  This trait is implemented by the various types of network messages.  Then instead of calling `receive()`, we call e.g. `receive::<RepairMessage>()` and do not have to write additional looping code.

The only interesting part of the RFC is in `src/network.rs`.  You can ignore the rest of the changes.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -155,6 +155,10 @@ pub enum NetworkReceiveError {
     BadSocket(#[from] std::io::Error),
 }
 
+pub trait FromNetworkMessage<T = Self> {
+    fn convert(src: NetworkMessage) -> Option<T>;
+}
+
 /// Abstraction of a network interface for sending and receiving messages.
 #[async_trait]
 pub trait Network: Send + Sync {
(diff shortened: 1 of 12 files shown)


src/network.rs
```diff
@@ -155,6 +155,10 @@ pub enum NetworkReceiveError {
     BadSocket(#[from] std::io::Error),
 }
 
+pub trait FromNetworkMessage<T = Self> {
```
qkniep, line 158: Seems almost like manually reimplementing `TryFrom`

qkniep:
I'm not sure how this would help. I also don't know what the purpose would be to try to receive different message types, if we receive them not in the expected order all calls to `receive` will fail. 

And then you give the callers of `send` and `receive` a lot of responsibility: If I want to call receive on e.g. `repair_network`, I first have to figure out that I have to call `repair_network.receive::<RepairResponse>()`, and if anyone ever calls `send` with the wrong type this results in an error that will only be caught by `receive` or its caller.

Issue #85, opened 2025-08-11
call `try_skip_window()` if we see invalid slice in `try_reconstruct_slice`


akhi3030:
Exactly, what we can even do is treat it the same as equivocation right now, so tigger `try_skip_window()` to vote to skip the leader

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/pull/82#discussion_r2266002166_

akhi3030:
Summarising the discussions from https://github.com/qkniep/alpenglow/pull/220#issuecomment-3339120648.

Adding a shred can return the following results:
- FirstShred, Block: generating the appropriate VotorEvent
- Ok or Duplicate: these are benign
- Equivocation and InvalidShred: 
  - These indicate leader misbehaviour.  We must not store any disseminated shreds afterwards.
  - We should rename the `equivocated` field to `leader_misbehaviour`
  - We want to keep the errors separate as in future we can slash for equivocation.
  - We can cal `try_skip_window()` but that is not strictly necessary.  Since we will not be accepting new shreds so the timeout will fire and call it.
- InvalidProof and InvalidSignature:
  - We cannot proof whether it was a misbehaving leader or relayer.  So we simply drop the shred.  We can combine these into `SignatureMismatch`.

PR #178, opened 2025-09-05
feat: propagate errors from slice and block reconstruction

akhi3030:
Part of https://github.com/qkniep/alpenglow/issues/85.

If `try_reconstruct_slice()` and `try_reconstruct_block()` fail, then propagate the error up the call stack to eventually call `try_skip_window()`.

In order to accomplish the above in a clean manner, the following changes are made:

- Introduces `ValidatedShred` which can only be constructed by verifying a shred.
- refactors `check_shred_to_add` and `add_valid_shred` so that the first function (`add_shred`) validates the shred and the second function (`add_validated_shred`) performs various checks, adds the shred, and calls `try_reconstruct_slice()` and `try_reconstruct_block()` as needed.

## Future work

- [ ] store and use `ValidatedShred` in more places

### DIFF
--- a/src/consensus/blockstore/slot_block_data.rs
+++ b/src/consensus/blockstore/slot_block_data.rs
@@ -171,6 +171,18 @@ impl BlockData {
         }
     }
 
+    fn add_shred(
+        &mut self,
+        shred: Shred,
+        leader_pk: PublicKey,
+    ) -> Result<Option<VotorEvent>, AddShredError> {
+        assert!(shred.payload().header.slot == self.slot);
+        let slice_index = shred.payload().header.slice_index;
+        let cached_merkle_root = self.merkle_root_cache.entry(slice_index);
+        let validated_shred = ValidatedShred::new(shred, cached_merkle_root, &leader_pk)?;
+        self.add_validated_shred(validated_shred)
+    }
+
     fn add_validated_shred(
         &mut self,
         validated_shred: ValidatedShred,
@@ -234,18 +246,6 @@ impl BlockData {
                 })),
             },
         }
-    }
-
-    fn add_shred(
-        &mut self,
-        shred: Shred,
-        leader_pk: PublicKey,
-    ) -> Result<Option<VotorEvent>, AddShredError> {
-        assert!(shred.payload().header.slot == self.slot);
-        let slice_index = shred.payload().header.slice_index;
-        let validated_shred =
-            ValidatedShred::new(shred, self.merkle_root_cache.entry(slice_index), &leader_pk)?;
-        self.add_validated_shred(validated_shred)
     }
 
     /// Reconstructs the slice if the blockstore contains enough shreds.
--- a/src/consensus/blockstore/slot_block_data.rs
+++ b/src/consensus/blockstore/slot_block_data.rs
@@ -15,8 +15,9 @@ use super::BlockInfo;
 use crate::consensus::votor::VotorEvent;
 use crate::crypto::signature::PublicKey;
 use crate::crypto::{Hash, MerkleTree};
-use crate::shredder::validated_shred::{ShredVerifyError, ValidatedShred};
-use crate::shredder::{DeshredError, RegularShredder, Shred, Shredder};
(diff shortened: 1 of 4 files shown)


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -28,8 +29,23 @@ pub enum AddShredError {
     Duplicate,
     #[error("shred shows leader equivocation")]
     Equivocation,
-    #[error("shred is after slice marked as last")]
-    AfterLastSlice,
+    #[error("invalid slice index relative to last slice")]
+    InvalidSliceIndex,
```
qkniep: not sure if we need these last error types (InvalidSliceIndex, ReconstructSlice, ReconstructBlock) as separate types, they all basically say "there is at least one shred that is invalid in a way that does not imply equivocation", no caller should ever care which of the three it is, the only reason I can see is for debugging

src/shredder/validated_shred.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Using the new type pattern, this struct provides the guarantee that various verification checks have been performaned on the encapsulated [`Shred`].
+#[derive(Debug, Clone)]
+#[repr(transparent)]
+pub struct ValidatedShred(Shred);
+
+impl ValidatedShred {
+    /// Performs various verification checks on the [`Shred`] and if they succeed, returns a shred.
+    ///
+    /// `cached_merkle_root`: is used to cache and skip certain checks for related shreds and detect equivocation.
+    pub fn new(
```
qkniep: should this be called `try_new()` because it returns a `Result`?

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -155,94 +171,83 @@ impl BlockData {
...
+        match (header.is_last, self.last_slice) {
+            (true, None) => {
+                self.last_slice = Some(slice_index);
+                self.slices.retain(|&ind, _| ind <= slice_index);
+                self.shreds.retain(|&ind, _| ind <= slice_index);
+            }
+            (true, Some(l)) => {
+                if slice_index != l {
+                    return Err(AddShredError::InvalidSliceIndex);
+                }
```
qkniep, line 200: could explicitly do this here:
```suggestion
                } else {
                    return Err(AddShredError::Duplicate);
                }
```

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -155,94 +171,83 @@ impl BlockData {
...
         }
 
-        // maybe send first shred notification
+        let is_first_shred = self.shreds.is_empty();
+        let slice_shreds = {
+            let shred_index = shred.payload().index_in_slice;
+            let slice_shreds = self.shreds.entry(slice_index).or_default();
+            let exists = slice_shreds
+                .iter()
+                .any(|s| s.payload().index_in_slice == shred_index);
```
qkniep, line 216: this always kind of bothers me, maybe it should rather be that `slice_shreds` holds something like `[Option<Shred>; TOTAL_SHREDS]`, maybe as a new PR...

qkniep (approved):
LGTM, made some minor comments

src/consensus/blockstore/slot_block_data.rs
akhi3030: makes sense.  combining them to `InvalidShred`.

src/consensus/blockstore/slot_block_data.rs
akhi3030: Are you sure this is a valid suggestion?  We are comparing `SliceIndex` here not `index_in_slice`.  One slice can have multiple shreds.  So we would end up discarding all but one shred for the last slice of the block.

qkniep: you are right

src/consensus/blockstore/slot_block_data.rs
akhi3030: https://github.com/qkniep/alpenglow/issues/182

PR #179, opened 2025-09-05
refactor: Pass the btreemap `Entry` to `Shred::verify()` to consolidate updating the cache and error checking

akhi3030:
If we pass the cache btreemap `Entry` to `Shred::verify()`, it can update it when the checks pass, use it to skip some checks, and we consolidate some error checking.

### DIFF
--- a/src/shredder.rs
+++ b/src/shredder.rs
@@ -19,6 +19,8 @@
 
 mod reed_solomon;
 
+use std::collections::btree_map::Entry;
+
 use aes::Aes128;
 use aes::cipher::{Array, KeyIvInit, StreamCipher};
 use ctr::Ctr64LE;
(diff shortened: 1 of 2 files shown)


src/shredder.rs
```diff
@@ -124,20 +137,39 @@ pub struct Shred {
...
         }
-        if Some(&self.merkle_root) == cached_merkle_root {
-            return true;
+
+        match cached_merkle_root {
+            Entry::Occupied(entry) => {
+                if entry.get() == &self.merkle_root {
+                    Ok(())
+                } else {
+                    Err(ShredVerifyError::Equivocation)
```
qkniep, line 163: as is, this would throw `ShredVerifyError::Equivocation` even for invalid signatures!
```suggestion
                    if self.merkle_root_sig.verify(&self.merkle_root, pk) {
                        Err(ShredVerifyError::Equivocation)
                    } else {
                        Err(ShredVerifyError::InvalidSignature)
                    }
```

akhi3030: Oh, that was subtle.  Good catch.  I couldn't come up with a cleaner way than what you suggested.  But given the subtlety, I decided to add some tests for the different paths here as well.

qkniep: thanks, the tests help

PR #180, opened 2025-09-09
ci: use sccache to cache build files

qkniep:


qkniep:
total pure build time decreased (only) from 8:30 to 5:30, despite sccache reported 100% cache hit rate...

qkniep:
ok, caching was not correctly enabled for code coverage builds, now down to 4:00 for all builds

qkniep:
ok, with both enabled we are back down to ~7 minutes for everything, 3:00 for all builds, not sure if we can do better, but it's still a >64% decrease in time spent on builds compared to `main`

.github/workflows/rust.yml
```diff
@@ -8,27 +8,36 @@ on:
...
       
-      - name: Cache build files
-        id: cache-build
-        uses: actions/cache@v4
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
```
akhi3030: I don't know what `override` does here.  Not sure if it is worth adding a comment to explain it.

akhi3030 (approved):
LGTM.

.github/workflows/rust.yml
qkniep: deleted the entire explicit install of stable Rust, which is already installed by default

PR #183, opened 2025-09-10
refactor: constrain what types of messages `Network` can handle

akhi3030:
Also see https://github.com/qkniep/alpenglow/pull/177 for more discussions.

High level idea is that consumers of `Network::receive()` have to sometimes handle message types that they are not interested in.  They generally handle this by ignoring them and continuing to loop.  We already have a similar loop in the structs that implement `Network` so the extra looping is unergonomic.

This patch adds constraints to `Network` that allows one to specify what types of messages they are willing to send and receive.  This allows the consumers to specify what types of messages they are interested in and also eliminate the extra looping.

Potential future work being tracked in https://github.com/qkniep/alpenglow/issues/187.

### DIFF
--- a/src/all2all/robust.rs
+++ b/src/all2all/robust.rs
@@ -6,6 +6,8 @@
 //! Broadcasts each message over the underlying instance of [`Network`].
 //! The message may be retransmitted multiple times.
 
+use async_trait::async_trait;
+
 use super::All2All;
 use crate::ValidatorInfo;
 use crate::network::{Network, NetworkMessage, NetworkReceiveError, NetworkSendError};
@@ -32,7 +34,8 @@ impl<N: Network> RobustAll2All<N> {
     pub fn handle_retransmits(&self) {}
 }
 
-impl<N> All2All for RobustAll2All<N>
+#[async_trait]
+impl<N: Network> All2All for RobustAll2All<N>
 where
     N: Network<Recv = NetworkMessage, Send = NetworkMessage>,
 {
--- a/src/all2all.rs
+++ b/src/all2all.rs
@@ -35,11 +35,14 @@
 mod robust;
 mod trivial;
 
+use async_trait::async_trait;
+
 pub use self::robust::RobustAll2All;
 pub use self::trivial::TrivialAll2All;
 use crate::network::{NetworkMessage, NetworkReceiveError, NetworkSendError};
 
 /// Abstraction for a direct all-to-all network communication protocol.
+#[async_trait]
 pub trait All2All {
     /// Broadcasts the given message to all known nodes.
     ///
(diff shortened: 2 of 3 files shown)


src/consensus/votor.rs
```diff
@@ -369,7 +369,7 @@ mod tests {
     use crate::network::{NetworkMessage, SimulatedNetwork};
     use crate::test_utils::{generate_all2all_instances, generate_validators};
 
-    type A2A = TrivialAll2All<SimulatedNetwork>;
+    type A2A = TrivialAll2All<SimulatedNetwork<NetworkMessage, NetworkMessage>>;
```
qkniep, line 372: This should probably use something like `ConsensusMessage` now, which can only be either `Vote` or `Cert`.

qkniep: Also, we can then fully abandon `NetworkMessage`

src/disseminator/rotor.rs
```diff
@@ -57,7 +57,10 @@ impl<N: Network> Rotor<N, FaitAccompli1Sampler<PartitionSampler>> {
     }
 }
 
-impl<N: Network, S: SamplingStrategy> Rotor<N, S> {
+impl<N, S: SamplingStrategy> Rotor<N, S>
+where
+    N: Network<Recv = NetworkMessage, Send = NetworkMessage>,
```
qkniep, line 62: This could just use `Shred` instead of the generic `NetworkMessage` we want to get rid of

qkniep: It could be useful to define this as a trait alias (e.g. "DisseminatorNetwork"). However, these are currently nightly only. What we could do in stable Rust is define a new (marker) trait for this.

src/all2all/robust.rs
```diff
@@ -32,7 +32,10 @@ impl<N: Network> RobustAll2All<N> {
     pub fn handle_retransmits(&self) {}
 }
 
-impl<N: Network> All2All for RobustAll2All<N> {
+impl<N> All2All for RobustAll2All<N>
+where
+    N: Network<Recv = NetworkMessage, Send = NetworkMessage>,
```
qkniep, line 40: if we end up specifying the associated types like this, maybe going to generic types would be more straightforward after-all 🙈

src/consensus/votor.rs
akhi3030: both should be tracked in https://github.com/qkniep/alpenglow/issues/187

src/disseminator/rotor.rs
akhi3030: Both should be tracked in https://github.com/qkniep/alpenglow/issues/187

src/all2all/robust.rs
akhi3030: Yeah I also wondered about this while preparing the PR.  I am still hoping that this is a cleaner longer term solution.

PR #185, opened 2025-09-11
feat: implement MCP simulations (Ryse + Pyjama)

qkniep:
Adds two of the drafts for MCP protocols, Ryse and Pyjama, to the `simulations` binary/module. Implements latency simulations for both based on the new discrete-event simulator, similar to the ones we already have for Rotor and Alpenglow. We also have some initial robustness calculations based on the parameter set and Binomial distribution calculations, thus ignoring the specifics of stake distribution and sampling strategies for now.

Builds on #228.


akhi3030 (approved):
Approving to unblock merging.  Let me know how detailed a review you want.

PR #186, opened 2025-09-11
refactor: use validated shreds in more places

akhi3030:
Part of #85

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -358,7 +358,7 @@ where
                 .blockstore
                 .write()
                 .await
-                .add_shred_from_disseminator(s)
+                .add_shred_from_disseminator(s.into_shred())
                 .await;
             if let Ok(Some((block_slot, block_info))) = block {
                 assert_eq!(block_slot, slot);
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -18,8 +18,7 @@ use super::epoch_info::EpochInfo;
 use super::votor::VotorEvent;
 use crate::consensus::blockstore::slot_block_data::BlockData;
 use crate::crypto::Hash;
-use crate::shredder::Shred;
-use crate::shredder::ValidatedShred;
+use crate::shredder::{Shred, ValidatedShred};
 use crate::types::SliceIndex;
 use crate::{Block, BlockId, Slot};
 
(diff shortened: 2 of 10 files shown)


src/consensus/blockstore.rs
```diff
@@ -301,12 +302,12 @@ impl Blockstore for BlockstoreImpl {
...
         slice_index: SliceIndex,
         shred_index: usize,
-    ) -> Option<&Shred> {
+    ) -> Option<&ValidatedShred> {
         let block_data = self.get_block_data(block_id)?;
         let slice_shreds = block_data.shreds.get(&slice_index)?;
         slice_shreds
             .iter()
-            .find(|s| s.payload().index_in_slice == shred_index)
+            .find(|s| s.to_shred().payload().index_in_slice == shred_index)
```
qkniep: Maybe we should just implement `AsRef<Shred>` on `ValidatedShred`?

src/shredder/validated_shred.rs
```diff
@@ -66,6 +66,10 @@ impl ValidatedShred {
         }
     }
 
+    pub(super) fn new_validated(shred: Shred) -> Self {
```
qkniep, line 73: missing docstring

src/consensus/blockstore.rs
akhi3030: In such scenarios, generally I have implemented `AsRef` for when I wanted to write functions that could accept `Shred` or `ValidatedShred`.  And without the `AsRef` implementation, there would be additional annoyance to implement such functions.  

Looking at our code, I can imagine that we could potentially write: `fn reed_solomon_deshred<S: AsRef<Shred>>(...)` if we wanted the ability to deshred both validated and non-validated shreds.  Although, I am not sure if we want to deshred a non-validated shred.  I don't see other use cases.

Could you please expand on why `AsRef` would be useful?

qkniep: oh, sorry for the confusion, I meant the `Deref` trait, so we could call `s.payload()` directly on `ValidatedShred`

akhi3030: Yeah, `Deref` can be useful here.  In general, I am wary of using "operator overloading" as there can be arbitrary things happening behind the scene and it is not even straightforward to figure out what code is executing.  But I think as long as it is a simple look up of the inner field, it should be fine.  I'll do it in a separate PR though.

src/consensus/block_producer.rs
```diff
@@ -358,7 +358,7 @@ where
                 .blockstore
                 .write()
                 .await
-                .add_shred_from_disseminator(s)
+                .add_shred_from_disseminator(s.into_shred())
```
qkniep, line 361: Here the leader turns a `ValidatedShred` into a `Shred` just to then validate it again within `add_shred_from_disseminator`. Instead, we could probably give the leader a shortcut to `blockstore.disseminated.add_validated_shred()`.

akhi3030: Agreed, created #216

Issue #182, opened 2025-09-10
Consider storing `BlockData::shreds` in a way to enable faster lookups


akhi3030:
this always kind of bothers me, maybe it should rather be that `slice_shreds` holds something like `[Option<Shred>; TOTAL_SHREDS]`, maybe as a new PR...

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/pull/178#discussion_r2336997080_
            

Relevant code is https://github.com/qkniep/alpenglow/blob/a229670683af91c66e664e52a155025cc107075f/src/consensus/blockstore/slot_block_data.rs#L229.  We are having to iterate over all the shreds to check for duplicates.  If we stored the shreds in an array or something similar, we could make this check faster.

qkniep:
Also since #166, we fill in all missing shreds after reconstruction, so this is not really wasting space compared to a `Vec`.

PR #190, opened 2025-09-12
refactor: store `BlockData::shreds` as arrays

akhi3030:
Closes: #182 

This PR ended up being much bigger than I originally hoped.

The key change that triggered the whole PR is to store `BlockData::shreds` in an array of size `TOTAL_SHREDS`.  

In order to make sure that we can index safely into the array and not have to handle out of bounds errors, I then introduced `ShredIndex` type to bound how we index into the array and used it in appropriate places.

This caused some places where we construct `ShredPayload` to use the new `ShredIndex` and either return errors or use `unwrap`s if we can guarantee that the number of shreds will be bound by `TOTAL_SHREDS`.  This requirement then caused me to change various interfaces in `shredders.rs` and `reed_solomon.rs` to use arrays instead of `Vec`.

### DIFF
--- a/src/shredder.rs
+++ b/src/shredder.rs
@@ -254,9 +254,9 @@ pub trait Shredder {
 pub struct RegularShredder;
 
 impl Shredder for RegularShredder {
-    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS;
-    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
     const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE;
+    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
+    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS;
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
@@ -299,9 +299,9 @@ impl Shredder for RegularShredder {
 pub struct CodingOnlyShredder;
 
 impl Shredder for CodingOnlyShredder {
-    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS;
-    const DATA_OUTPUT_SHREDS: usize = 0;
     const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE;
+    const DATA_OUTPUT_SHREDS: usize = 0;
+    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS;
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
@@ -348,10 +348,10 @@ impl Shredder for CodingOnlyShredder {
 pub struct PetsShredder;
 
 impl Shredder for PetsShredder {
-    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS + 1;
-    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS - 1;
     // needs 16 bytes for symmmetric encryption key
     const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE - 16;
+    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS - 1;
+    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS + 1;
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
@@ -430,10 +430,10 @@ impl Shredder for PetsShredder {
 pub struct AontShredder;
 
 impl Shredder for AontShredder {
-    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS;
-    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
     // needs 16 bytes for symmmetric encryption key
     const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE - 16;
+    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
+    const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS;
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -3,5 +3,4 @@ unstable_features = true
 
 group_imports = "StdExternalCrate"
 imports_granularity = "Module"
-reorder_impl_items = true
 use_field_init_shorthand = true


benches/shredder.rs
```diff
@@ -41,9 +41,13 @@ fn deshred<S: Shredder>(bencher: divan::Bencher) {
             let slice = create_slice_with_invalid_txs(size);
             let mut rng = rand::rng();
             let sk = SecretKey::new(&mut rng);
-            S::shred(slice, &sk).unwrap()
+            let mut shreds = S::shred(slice, &sk).unwrap().map(Some);
+            for shred in shreds.iter_mut().skip(DATA_SHREDS) {
```
qkniep: the equivalent to the old code would be to have `take(DATA_SHREDS)` here, right? because it's the shreds we are overwriting with `None`, i.e. not passing to `deshred()`

but also in the old code this is wrong because it implicitly assumes `TOTAL_SHREDS = 2 * DATA_SHREDS` (or at the very least `TOTAL_SHREDS >= 2 * DATA_SHREDS`), neither of which need generally be true, thus we should really have something like `take(TOTAL_SHREDS - DATA_SHREDS)` here, so that we only pass `DATA_SHREDS` many shreds (from the end, so likely coding shreds) to `deshred()`

src/shredder/validated_shred.rs
```diff
@@ -102,7 +102,7 @@ mod tests {
     fn create_random_shred() -> (Shred, SecretKey) {
         let sk = SecretKey::new(&mut rng());
         let slice = create_slice_with_invalid_txs(MAX_DATA_PER_SLICE - 16);
-        let mut shreds = RegularShredder::shred(slice, &sk).unwrap();
+        let mut shreds = RegularShredder::shred(slice, &sk).unwrap().to_vec();
```
qkniep: it's only used for testing, so maybe it doesn't matter, but cloning the specific shred seems like a better idea than allocating an entire `Vec`

src/shredder.rs
```diff
@@ -556,6 +587,15 @@ mod tests {
     use super::*;
     use crate::types::slice::create_slice_with_invalid_txs;
 
+    fn into_array(shreds: &[ValidatedShred]) -> [Option<ValidatedShred>; TOTAL_SHREDS] {
```
qkniep, line 637: this only works if `shreds` contains a prefix of all shreds, is this intended? a comment about this would be good

src/shredder.rs
```diff
@@ -564,39 +604,47 @@ mod tests {
         assert_eq!(shreds.len(), TOTAL_SHREDS);
 
         // restore from all shreds
-        let (slice_restored, _) = RegularShredder::deshred(&shreds)?;
+        let all = into_array(&shreds);
+        let (slice_restored, _) = RegularShredder::deshred(&all)?;
         slice.merkle_root = slice_restored.merkle_root;
         assert_eq!(slice_restored, slice);
 
         // restore only from coding shreds
```
qkniep: oops, this is actually data?

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
         .collect();
 
     Ok((data_shreds, coding_shreds))
 }
 
 /// Reconstructs the raw data from the given shreds.
+///
+/// Errors
+///
+/// If fewer than DATA_SHREDS elements in `shreds` are Some() then returns Err(ReedSolomonDeshredError::NotEnoughShreds).
```
qkniep:
```suggestion
/// If fewer than [`DATA_SHREDS`] elements in `shreds` are `Some()` then returns `Err(ReedSolomonDeshredError::NotEnoughShreds)`.
```

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
+/// If the restored payload is larger than [`MAX_DATA_PER_SLICE_AFTER_PADDING`] then returns Err(ReedSolomonDeshredError::TooMuchData).
```
qkniep:
```suggestion
/// If the restored payload is larger than [`MAX_DATA_PER_SLICE_AFTER_PADDING`] then returns `Err(ReedSolomonDeshredError::TooMuchData)`.
```

src/shredder.rs
```diff
@@ -564,39 +604,47 @@ mod tests {
...
-        let (slice_restored, _) = RegularShredder::deshred(&shreds[..DATA_SHREDS])?;
+        let coding = into_array(&shreds[..DATA_SHREDS]);
+        let (slice_restored, _) = RegularShredder::deshred(&coding)?;
         assert_eq!(slice_restored, slice);
 
         // restore only from data shreds
```
qkniep: oops, this is actually coding?

src/shredder.rs
qkniep: or the point is that the array it outputs needs not be aligned with shred indexes, which is also peculiar and should be commented

src/shredder.rs
```diff
@@ -214,7 +213,9 @@ pub trait Shredder {
     ///     1. Reconstruct all shreds (data and coding) under the Merkle tree.
     ///     2. Verify the entire Merkle tree.
     ///     3. Return [`DeshredError::InvalidMerkleTree`] if this fails.
-    fn deshred(shreds: &[ValidatedShred]) -> Result<(Slice, Vec<ValidatedShred>), DeshredError>;
+    fn deshred(
```
qkniep, line 218: so `deshred` as of now takes an array but still needs to work correctly, even if the shreds' positions in the array do not line up with their `ShredIndex`es, which seem very counter-intuitive to me... either we should continue giving it a `Vec` or (probably better) it should be able to assume that it always gets an array where the indices line up

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
     };
+
+    // unwrap below should be safe because of the assertion above
     let data_shreds = payload
         .chunks(shred_bytes)
         .enumerate()
-        .map(|(i, p)| DataShred(payload_from_index_and_data(i, p.to_vec())))
+        .map(|(i, p)| {
+            DataShred(payload_from_index_and_data(
+                ShredIndex::new(i).unwrap(),
```
qkniep: this seems like a misuse of the `ShredIndex` type, because this is not yet the index in the slice but just a number from 0 to `num_data`, which depending on the `Shredder` may or may not map to some `ShredIndex` (e.g. for `CodingOnlyShredder`, all of these here would never end up in the final slice)

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
+                p.to_vec(),
+            ))
+        })
         .collect();
+    // unwrap below should be safe because of the assertion above
```
qkniep: the `unwrap()` below would require something like `assert(num_coding <= TOTAL_SHREDS)`, which we do not do above

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
     let coding_shreds = coding_parts
         .into_iter()
         .enumerate()
-        .map(|(i, p)| CodingShred(payload_from_index_and_data(i, p)))
+        .map(|(i, p)| CodingShred(payload_from_index_and_data(ShredIndex::new(i).unwrap(), p)))
```
qkniep: same here as above, this is not yet the final index in the slice

src/shredder/reed_solomon.rs
qkniep: it could still be valid, although a bit confusing maybe to continue using `ShredIndex` here as well, valid because it's sensible to assume that all shredders would never deal with more than `TOTAL_SHREDS` data shreds and `TOTAL_SHREDS` coding shreds (so the bounds for each of the two would hold, confusing because now `ShredIndex` again can mean two different things (index of an output shred within the `TOTAL_SHREDS` shreds of a slice, or index of a data/coding shred within the "however many" data/coding shreds the `Shredder` deals with)

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
     assert!(coding_offset <= DATA_SHREDS);
-    if shreds.len() < DATA_SHREDS {
+    let shreds_cnt = shreds.iter().filter(|s| s.is_some()).count();
+    if shreds_cnt < DATA_SHREDS {
         return Err(ReedSolomonDeshredError::NotEnoughShreds);
     }
-    if shreds.len() > TOTAL_SHREDS {
-        return Err(ReedSolomonDeshredError::TooManyShreds);
-    }
+    assert!(shreds.len() <= TOTAL_SHREDS);
```
qkniep: do we even need this? if anything this could be a `const_assert` on the type of shreds if that's possible b/c it's always true at compile time

src/shredder/reed_solomon.rs
qkniep: for all shredders we have so far this definitely works

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
 
 /// Reconstructs the raw data from the given shreds.
+///
+/// Errors
+///
+/// If fewer than DATA_SHREDS elements in `shreds` are Some() then returns Err(ReedSolomonDeshredError::NotEnoughShreds).
+/// If the restored payload is larger than [`MAX_DATA_PER_SLICE_AFTER_PADDING`] then returns Err(ReedSolomonDeshredError::TooMuchData).
 pub(super) fn reed_solomon_deshred(
-    shreds: &[ValidatedShred],
+    shreds: &[Option<ValidatedShred>; TOTAL_SHREDS],
```
qkniep, line 67: here also, we probably want to assume the indices always match

src/shredder/reed_solomon.rs
```diff
@@ -51,48 +48,65 @@ pub(super) fn reed_solomon_shred(
...
-    if shreds.len() > TOTAL_SHREDS {
-        return Err(ReedSolomonDeshredError::TooManyShreds);
-    }
+    assert!(shreds.len() <= TOTAL_SHREDS);
 
     // filter to split data and coding shreds
-    let data = shreds.iter().filter_map(|s| match &s.payload_type {
-        ShredPayloadType::Data(d) => Some((d.index_in_slice, &d.data)),
-        ShredPayloadType::Coding(_) => None,
+    let data = shreds.iter().filter_map(|s| {
```
qkniep: makes this easier as well, as we can just look at `shreds[..coding_offset]` and assert that they have type `ShredPayloadType::Data`, respectively below look at `shreds[coding_offset..]` and assert that they have type `ShredPayloadType::Coding`

src/shredder/reed_solomon.rs
```diff
@@ -218,10 +221,14 @@ mod tests {
...
         data: Vec<DataShred>,
         coding: Vec<CodingShred>,
-    ) -> Vec<ValidatedShred> {
+    ) -> [Option<ValidatedShred>; TOTAL_SHREDS] {
         let sk = SecretKey::new(&mut rand::rng());
         let shreds = data_and_coding_to_output_shreds(data, coding, &sk);
         // reverse order to get coding shreds, not just data shreds
-        shreds.into_iter().rev().take(DATA_SHREDS).collect()
+        let mut shreds = shreds.into_iter().rev().map(Some).collect::<Vec<_>>();
+        for shred in shreds.iter_mut().skip(DATA_SHREDS) {
```
qkniep: unnecessary to destroy the order of the array here: just `take(TOTAL_SHREDS - DATA_SHREDS)` for the loop and remove `rev()` above

qkniep: also do we need to collect into a `Vec` before turning into an array? (ah, this is only for testing again? then it doesn't matter as much)

src/shredder.rs
```diff
@@ -525,8 +558,9 @@ fn create_output_shreds_for_other_leader(
         shreds.push(ValidatedShred::new_validated(shred));
     }
     for mut c in coding {
-        c.0.index_in_slice += num_data_shreds;
-        let merkle_path = tree.create_proof(c.0.index_in_slice);
+        // based on assertion above, this unwrap() should be safe
```
qkniep: in addition to the assertion above (comparing to `TOTAL_SHREDS`) we implicitly assume that the `shred_index`es of the `CodingShred`s are `0..coding.len()`

qkniep: one way to possibly completely circumvent this entire problem, would be to just use the indices within `data` and `coding` instead, i.e. the shreds in `data` get `ShredIndex`es `0..data.len()` and the ones in `coding` get `data.len()..TOTAL_SHREDS`?!

qkniep (commented):
General APIs look great. However, I made quite a few comments. The most important ones are related to two main questions:
I think it would be great if everywhere where `[Option<ValidatedShred>; TOTAL_SHREDS]` is used, we could rely on the fact that a shred with `ShredIndex` `i` also appears at position `i` in that array. This is currently not the case. Most importantly, this appears in `deshred()` and what it can assume about the array order: https://github.com/qkniep/alpenglow/pull/190/files#r2376008759
The other one is related to a dual use of `ShredIndex` that is inherited from how the corresponding `usize`s were used before, see: https://github.com/qkniep/alpenglow/pull/190/files#r2376067497
and related to that: https://github.com/qkniep/alpenglow/pull/190/files#r2376207869

benches/shredder.rs
akhi3030: Yes, makes sense.  https://github.com/qkniep/alpenglow/pull/190/commits/e460f9010eac8c841e65a3222acfd5fb12d35444

src/shredder/validated_shred.rs
akhi3030: fixed in https://github.com/qkniep/alpenglow/pull/190/commits/7496e3af3ee4a5d12e3478a2d70c0d68f348b3ba.

src/shredder/reed_solomon.rs
akhi3030: Hm... I see the issue.  I think the problem is that we are reusing the `DataShred` and `CodingShred` types in this module where `ShredIndex` does not really have any real meaning.  It only starts to have meaning once it is constructed by the different shredders.  So maybe the right way to handle this would be to introduce some new types for this module.

akhi3030: I think I have an idea on how to address this.  Working on a separate PR for it.

akhi3030: https://github.com/qkniep/alpenglow/pull/221 should help with this issue.

src/shredder/reed_solomon.rs
akhi3030: `const_assert` doesn't work because `shreds` is not `const`.  I'll get rid of it.  It was mainly to serve as a readability tool.

src/shredder/reed_solomon.rs
akhi3030: https://github.com/qkniep/alpenglow/pull/190/commits/1dcd52a44f7afea39414b6038e486e4d2804ec95

src/shredder.rs
akhi3030: Now fixed.

src/shredder.rs
akhi3030: Added runtime checks for this.

src/shredder/reed_solomon.rs
akhi3030: added runtime checks.

src/shredder.rs
```diff
@@ -229,6 +244,21 @@ impl Shredder for RegularShredder {
     fn deshred(
         shreds: &[Option<ValidatedShred>; TOTAL_SHREDS],
     ) -> Result<(Slice, [ValidatedShred; TOTAL_SHREDS]), DeshredError> {
+        let () = validate_shred_indexes(&shreds)?;
+        for i in 0..DATA_SHREDS {
```
qkniep: could avoid a lot of duplication here by turning this into a function `validate_shred_layout` that takes something like `num_data` and `num_coding` and returns `Result<(), DeshredError>`

qkniep: could also subsume `validate_shred_indexes`

qkniep: also, a thought I just had: we might want to have these as `const`s on the `impl Shredder`, could call them something like `DATA_OUTPUT_SHREDS` and `CODING_OUTPUT_SHREDS`

qkniep: then validate_shred_layout also does not need to take them as arguments, can just be a provided method of the `Shredder` trait

akhi3030: good idea, now done.

src/shredder.rs
```diff
@@ -181,6 +182,12 @@ pub trait Shredder {
     /// However, this can be less if the specfic shredder adds some overhead.
     const MAX_DATA_SIZE: usize;
 
+    /// When [`deshred()`] is called, how many data shreds will be produced.
```
qkniep: I mean for `deshred()` this kind of applies as well because of the reconstruction, but that's really secondary
```suggestion
    /// When [`shred()`] is called, how many data shreds will be produced.
```

src/shredder.rs
```diff
@@ -181,6 +182,12 @@ pub trait Shredder {
...
+    const DATA_OUTPUT_SHREDS: usize;
+
+    /// When [`deshred()`] is called, how many coding shreds will be produced.
```
qkniep: same

src/shredder.rs
```diff
@@ -254,9 +254,9 @@ pub trait Shredder {
 pub struct RegularShredder;
 
 impl Shredder for RegularShredder {
-    const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE;
-    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
     const CODING_OUTPUT_SHREDS: usize = TOTAL_SHREDS - DATA_SHREDS;
+    const DATA_OUTPUT_SHREDS: usize = DATA_SHREDS;
+    const MAX_DATA_SIZE: usize = MAX_DATA_PER_SLICE;
```
qkniep, line 257: I liked the previous order way better, especially CODING before DATA, is very confusing, we do it the other way around everywhere else

qkniep: I see, I think the culprit is `reorder_impl_items`, in `rustfmt.toml`, should get rid of that

qkniep (approved):
LGTM now, thanks

Issue #187, opened 2025-09-11
Future work after constraining `Network` trait


akhi3030:
After https://github.com/qkniep/alpenglow/pull/183 lands, we should explore doing the following additional improvements:

PR #202, opened 2025-09-12
refactor: remove `RepairMessage`

akhi3030:
Part of #187 
Closes: #193 

At this point in time, `RepairMessage` is essentially a wrapper which is not particularly useful and we can use `RepairRequest` and `RepairResponse` directly.  This also makes it clearer which types of messages the different `Repair` related structs are handling.

### DIFF
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -392,8 +392,7 @@ where
             sender: self.epoch_info.own_id,
             req_type,
         };
-        let msg_bytes =
-            bincode::serde::encode_to_vec(request, bincode::config::standard()).unwrap();
+        let msg_bytes = bincode::serde::encode_to_vec(request, BINCODE_CONFIG).unwrap();
         // HACK: magic number to fix high-failure scenarios
         let mut to_all = HashSet::new();
         for _ in 0..10 {
--- a/benches/network.rs
+++ b/benches/network.rs
@@ -3,7 +3,7 @@
 
 use alpenglow::consensus::Vote;
 use alpenglow::crypto::{Hash, aggsig, signature};
-use alpenglow::network::NetworkMessage;
+use alpenglow::network::{BINCODE_CONFIG, NetworkMessage};
 use alpenglow::shredder::{MAX_DATA_PER_SLICE, RegularShredder, Shredder};
 use alpenglow::types::Slot;
 use alpenglow::types::slice::create_slice_with_invalid_txs;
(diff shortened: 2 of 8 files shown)


src/network.rs
```diff
@@ -58,7 +57,6 @@ pub enum NetworkMessage {
     Shred(Shred),
     Vote(Vote),
     Cert(Cert),
-    Repair(RepairMessage),
     // FIXME: txs should not be seen on the same connection as other network msgs.
     // This should not be part of this enum.
     Transaction(Transaction),
```
qkniep, line 61: can't we remove this one as well already?

src/repair.rs
```diff
@@ -392,23 +401,23 @@ where
...
         self.request_timeouts.push((expiry, hash));
 
-        let repair = RepairMessage::Request(request, self.epoch_info.own_id);
-        let msg_bytes = bincode::serde::encode_to_vec(repair, bincode::config::standard()).unwrap();
+        let request = RepairRequest {
+            sender: self.epoch_info.own_id,
+            req_type,
+        };
+        let msg_bytes =
+            bincode::serde::encode_to_vec(request, bincode::config::standard()).unwrap();
```
qkniep:
```suggestion
            bincode::serde::encode_to_vec(request, BINCODE_CONFIG).unwrap();
```

src/network.rs
akhi3030: Yeah, we could.   I tried but it actually impacts a few more things, some tests and the workload generator.  So I think it's better to handle it in a separate PR.  I don't think we need a separate issue for it.  #199 should cover it.

src/repair.rs
```diff
@@ -53,6 +44,27 @@ pub enum RepairRequest {
...
+        let msg_bytes = bincode::serde::encode_to_vec(repair, BINCODE_CONFIG).unwrap();
+        hash(&msg_bytes)
+    }
+}
+
+/// Request messages for the repair sub-protocol.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RepairRequest {
+    /// The validator that sent the message.
+    sender: ValidatorId,
```
qkniep, line 62: for now this is good, but having the sender in here was more of a workaround... ideally we would want to determine the sender at the network level, e.g. based on the UDP sender address, but more importantly for protocols with standing/authenticated connections such as TCP, TCP+TLS, QUIC based on the open connection and valid auth

src/repair.rs
akhi3030: fixed this and it all the other places too.

src/repair.rs
akhi3030: indeed... I also thought it was odd and figured that it is a shortterm thing.

PR #208, opened 2025-09-15
refactor: remove `Networkmessage`

akhi3030:
Part of #187 
Closes: #199


akhi3030:
Marking as draft.  I forgot to remove references in some files.

src/network/simulated/core.rs
```diff
@@ -254,21 +254,34 @@ impl Default for SimulatedNetworkCore {
...
 
     use super::*;
-    use crate::network::{Network, NetworkMessage, localhost_ip_sockaddr};
+    use crate::network::{Network, localhost_ip_sockaddr};
 
     // test simulated latency accuracy to within +/-5%
     const ACCURACY: f64 = 0.05;
 
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    struct Ping();
```
qkniep: is there any difference/advantage of `Ping()` over `Ping`?

akhi3030: Nope, changed `Ping`

src/test_utils.rs
```diff
@@ -17,6 +18,18 @@ use crate::{
     BlockId, MAX_TRANSACTION_SIZE, Slot, Transaction, ValidatorId, ValidatorInfo, VotorEvent,
 };
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Ping;
```
qkniep, line 22: do we need this, or would it be better to just us `PingOrPong` everywhere?

src/test_utils.rs
```diff
@@ -17,6 +18,18 @@ use crate::{
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Pong;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum PingOrPong {
```
qkniep, line 28: Do we even still need these at all? Could just use `String` or some primitive type as message type for testing.

akhi3030:
I guess I'll merge this PR then.  We could probably make improvements but as this is test code and not a lot of it, it isn't so important.

PR #209, opened 2025-09-18
refactor: simplify `Network` trait

akhi3030:
Part of #187 
Closes: #196 

As the network trait already supports sending serialized data, we can remove the ability to send non-serialized data from it.  This simplifies the trait.  However, the downside is that now components can send arbitrary messages.  But compared to status quo, it is no worse as they could already send arbitrary messages by using `send_serialized`.  

- Removes `send()` from the `Network` trait
- renames `send_serialized()` to `send()`

### DIFF
--- a/src/all2all/robust.rs
+++ b/src/all2all/robust.rs
@@ -11,7 +11,7 @@ use async_trait::async_trait;
 use super::All2All;
 use crate::ValidatorInfo;
 use crate::consensus::ConsensusMessage;
-use crate::network::Network;
+use crate::network::{Network, into_bytes};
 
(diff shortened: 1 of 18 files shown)


qkniep:
As we discussed today: What we should probably do instead is keep type-safety and instead make `send_serialized()` private. To allow for that the `Network` trait should allow for a way to send the same message to multiple addresses, only serializing the message once, e.g. provide a method `send_to_many(&self, msg: Self::Send, to: Vec<SocketAddr>)` or `send_to_many(&self, msg: Self::Send, to: impl Iterator<Item = SocketAddr>)`.

Issue #210, opened 2025-09-19
Remove `send_serialized()` from the public interface


akhi3030:
See https://github.com/qkniep/alpenglow/pull/209#issuecomment-3309623813 for more details.

PR #211, opened 2025-09-19
refactor: remove `send_serialized()` from `Network` trait

akhi3030:
Part of #187 
Closes of #210 

- Additionally allows us to a couple of opportunities of improvements where we can sent to multiple addresses concurrently.

### DIFF
--- a/src/all2all/robust.rs
+++ b/src/all2all/robust.rs
@@ -6,6 +6,8 @@
 //! Broadcasts each message over the underlying instance of [`Network`].
 //! The message may be retransmitted multiple times.
 
+use std::iter::repeat;
+
 use async_trait::async_trait;
 
 use super::All2All;
@@ -41,13 +43,12 @@ where
     N: Network<Recv = ConsensusMessage, Send = ConsensusMessage>,
 {
     async fn broadcast(&self, msg: &ConsensusMessage) -> std::io::Result<()> {
-        for v in &self.validators {
-            // HACK: stupidly expensive retransmits
-            for _ in 0..1000 {
-                self.network.send(msg, v.all2all_address).await?;
-            }
-        }
-        Ok(())
+        let addrs = self
+            .validators
+            .iter()
+            .map(|v| repeat(v.all2all_address).take(1000))
+            .flatten();
+        self.network.send(msg, addrs).await
     }
 
     async fn receive(&self) -> std::io::Result<ConsensusMessage> {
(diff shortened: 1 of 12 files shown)


src/all2all/robust.rs
```diff
@@ -41,13 +43,11 @@ where
     N: Network<Recv = ConsensusMessage, Send = ConsensusMessage>,
 {
     async fn broadcast(&self, msg: &ConsensusMessage) -> std::io::Result<()> {
-        for v in &self.validators {
-            // HACK: stupidly expensive retransmits
```
qkniep, line 45: We should probably keep this comment.

src/consensus/block_producer.rs
```diff
@@ -562,7 +563,7 @@ mod tests {
             for i in 0..255 {
                 let data = vec![i; MAX_TRANSACTION_SIZE];
                 let msg = Transaction(data);
-                txs_sender.send(&msg, addr).await.unwrap();
+                txs_sender.send(&msg, once(addr)).await.unwrap();
```
qkniep, line 566: You think it's better to have just the `Iterator`-based `send()` method and no separate `send_to_many()`?

src/network.rs
```diff
@@ -47,9 +47,17 @@ pub trait Network: Send + Sync {
...
 
-    async fn send(&self, message: &Self::Send, to: SocketAddr) -> std::io::Result<()>;
-
-    async fn send_serialized(&self, bytes: &[u8], to: SocketAddr) -> std::io::Result<()>;
+    /// Sends the `message` to all the addresses in `addrs`.
+    ///
+    /// Note that a possible strategy for the implementators is to send to one address after another.
+    /// In this strategy, it is possible that if sending to one address fails, the implementator gives up sending to the remaining addresses.
+    ///
+    /// NOTE: Consider return a `Vec<Result<()>>` to indicate per address failures.
```
qkniep:
```suggestion
    //
    // NOTE: Consider return a `Vec<Result<()>>` to indicate per address failures.
```

qkniep (approved):

src/consensus/block_producer.rs
akhi3030: I am hoping that the additional cost of constructing the iterator will probably only add a few cycles and not have a significant impact compared to doing a syscall.  If and when we start doing some batching or use `io_uring` then I think it will become more interesting.

qkniep: I didn't even mean performance overhead at all, just ergonomics.

akhi3030: yeah, I can see that.  I will work on a follow up PR.

PR #212, opened 2025-09-22
refactor: use marker traits for different variants of `Network`

### LINKED ISSUE #200: Explore using marker traits (https://github.com/qkniep/alpenglow/pull/183/files#r2341498427)

akhi3030:
Part of #187 
Closes #200 

Introduces some new traits that constrain the `Network` trait and uses them in various places.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -35,6 +35,10 @@ use async_trait::async_trait;
 pub use self::simulated::SimulatedNetwork;
 pub use self::tcp::TcpNetwork;
 pub use self::udp::UdpNetwork;
+use crate::Transaction;
+use crate::consensus::ConsensusMessage;
+use crate::repair::{RepairRequest, RepairResponse};
+use crate::shredder::Shred;
 
 /// Maximum payload size of a UDP packet.
 pub const MTU_BYTES: usize = 1500;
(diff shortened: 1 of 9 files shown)


src/network.rs
```diff
@@ -64,6 +68,31 @@ pub trait Network: Send + Sync {
     async fn receive(&self) -> std::io::Result<Self::Recv>;
 }
 
+/// A marker trait that constrains [`Network`] to send and receive [`Shred`]
+pub trait ShredNetwork: Network<Recv = Shred, Send = Shred> {}
+impl ShredNetwork for UdpNetwork<Shred, Shred> {}
+impl ShredNetwork for SimulatedNetwork<Shred, Shred> {}
```
qkniep: should we use blanket implementations, like this instead?
```rust
impl<N> ShredNetwork for N where N: Network<Recv = Shred, Send = Shred> {}
```

qkniep (approved):
LGTM, I was just thinking that providing blanket implementations for these traits makes more sense. What do you think?

src/network.rs
akhi3030: thanks, much cleaner.

PR #213, opened 2025-09-22
refactor: introduce ability to send to a single address

akhi3030:
Also see https://github.com/qkniep/alpenglow/pull/211#discussion_r2361612892.

To make things more ergonomic:

- Changes `send` to that it can only send to a single address
- introduces `send_to_many` to allow sending to many addresses

### DIFF
--- a/src/all2all/robust.rs
+++ b/src/all2all/robust.rs
@@ -48,7 +48,7 @@ where
             .validators
             .iter()
             .flat_map(|v| repeat_n(v.all2all_address, 1000));
-        self.network.send(msg, addrs).await
+        self.network.send_to_many(msg, addrs).await
     }
 
     async fn receive(&self) -> std::io::Result<ConsensusMessage> {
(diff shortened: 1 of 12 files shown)


akhi3030:
@qkniep: check out the second commit.  If I do not send sequentially in simulated network, there are some errors and some tests were failing.  I didn't debug this in depth in this PR.  Is this a known issue?  Is this something that is worth investigating?

qkniep:
that is indeed strange behavior, I looked at it briefly, the tests don't appear like they should depend on the order of messages the `broadcast()` call sends them in under the hood... however, I was able to recreate the same behavior locally `simple_broadcast` and `packet_loss` don't terminate... interestingly enough `extreme_packetloss` (the same test with 90% packet loss instead of 0% or 20%) runs successfully

akhi3030:
I think https://github.com/qkniep/alpenglow/pull/215 should fix the problem.

PR #215, opened 2025-09-22
bug: fix some broadcast test failures

akhi3030:
In https://github.com/qkniep/alpenglow/pull/213, we observed some test failures.  Debugging it, I believe the following is happening:

- The sender is broadcasting and then exiting without receiving the messages that it broadcast
- if we send the messages too quickly as is happening in #213, the sender is exiting before simulated/core is able to deliver the msgs to it.
- the sender exits and then the core tries to send it message, finds that the channel is closed, and panics
- now the other receivers get stuck as they have not received all of the messages they wanted

The proposed fix is for the sender to receive messages that it broadcast before exiting and for the the core to not panic if it fails deliver messages.

### DIFF
--- a/src/network/simulated/core.rs
+++ b/src/network/simulated/core.rs
@@ -82,6 +82,7 @@ impl SimulatedNetworkCore {
                     let n_guard = n.read().await;
                     let channel = n_guard.get(&msg.to).unwrap();
                     if let Err(_e) = channel.send(msg).await {
+                        #[cfg(test)]
                         println!("sending failed. Ignoring");
                         warn!("sending failed. Ignoring");
                     }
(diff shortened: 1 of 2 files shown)


src/network/simulated/core.rs
```diff
@@ -80,7 +81,10 @@ impl SimulatedNetworkCore {
                     let msg = guard.pop().unwrap();
                     let n_guard = n.read().await;
                     let channel = n_guard.get(&msg.to).unwrap();
-                    channel.send(msg).await.unwrap();
+                    if let Err(_e) = channel.send(msg).await {
+                        println!("sending failed. Ignoring");
```
akhi3030, line 86: using print so that it shows up in tests without having to setup logging which is tricky to do.

qkniep, line 86: should we only enable this in tests then? to avoid noisy output
```suggestion
                        #[cfg(test)]
                        println!("sending failed. Ignoring");
```

akhi3030: makes sense.

PR #220, opened 2025-09-24
wip: feat: handle different types of leader misbehavior better

akhi3030:
Part of #85 

- spells out the different scenarios where the leader provably misbehaved and where we got a faulty shred without proof for for who misbehaved
- stops accepting new disseminated shreds when the leader misbehaved
- Improvements to API

### DIFF
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -43,12 +43,8 @@ impl From<&Block> for BlockInfo {
 #[async_trait]
 #[automock]
 pub trait Blockstore {
-    async fn add_shred_from_disseminator(&mut self, shred: Shred) -> Option<(Slot, BlockInfo)>;
-    async fn add_shred_from_repair(
-        &mut self,
-        hash: Hash,
-        shred: Shred,
-    ) -> Option<(Slot, BlockInfo)>;
+    async fn add_shred_from_disseminator(&mut self, shred: Shred) -> Option<BlockInfo>;
+    async fn add_shred_from_repair(&mut self, hash: Hash, shred: Shred) -> Option<BlockInfo>;
     fn disseminated_block_hash(&self, slot: Slot) -> Option<Hash>;
     #[allow(clippy::needless_lifetimes)]
     fn get_block<'a>(&'a self, block_id: BlockId) -> Option<&'a Block>;
@@ -99,11 +95,10 @@ impl BlockstoreImpl {
         self.block_data = self.block_data.split_off(&slot);
     }
 
-    async fn handle_add_shred_result(
-        &self,
-        slot: Slot,
-        res: AddShredResult,
-    ) -> Option<(Slot, BlockInfo)> {
+    /// Handles the result of attempt to insert a [`Shred`] into the [`SlotBlockData`].
+    ///
+    /// Generates appropriate [`VotorEvent`]s and returns [`Some(BlockInfo)`] if a block was reconstructed.
+    async fn handle_add_shred_result(&self, slot: Slot, res: AddShredResult) -> Option<BlockInfo> {
         match res {
             AddShredResult::FirstShred => {
                 self.votor_channel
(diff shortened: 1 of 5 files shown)


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -58,6 +32,18 @@ pub struct SlotBlockData {
...
+#[derive(Debug)]
+pub(super) enum AddShredResult {
+    FirstShred(Slot),
+    Block { slot: Slot, block_info: BlockInfo },
+    Equivocation(Slot),
+    Ok,
+    InvalidProof,
+    InvalidSignature,
+    InvalidShred,
+    Duplicate,
```
akhi3030: We don't really handle different error scenarios separately.  So I propose that we just return a `Error` here.

qkniep:
yes, will look at this next, currently reviewing #190 which takes some time

src/consensus/blockstore.rs
```diff
@@ -103,25 +99,41 @@ impl BlockstoreImpl {
...
+            AddShredResult::FirstShred(slot) => {
+                self.votor_channel
+                    .send(VotorEvent::FirstShred(slot))
+                    .await
+                    .unwrap();
+                None
+            }
+            AddShredResult::Equivocation(slot) => {
+                self.votor_channel
+                    .send(VotorEvent::SafeToSkip(slot))
```
qkniep: this is very unsafe! safe-to-skip says that we saw enough votes from others to know that no block was fast finalized in the slot and thus it's valid to vote skip-fallback on it...

src/consensus/blockstore.rs
```diff
@@ -103,25 +99,41 @@ impl BlockstoreImpl {
...
                     &hex::encode(block_info.parent.1)[..8],
                     block_info.parent.0,
                 );
-
                 Some((slot, block_info))
             }
-            ev => panic!("unexpected event {ev:?}"),
+            AddShredResult::Ok => None,
+            rest => {
+                warn!("Adding shred failed with {rest:?}");
```
qkniep:
```suggestion
                warn!("adding shred failed with: {rest:?}");
```

src/consensus/blockstore/slot_block_data.rs
qkniep: `InvalidShred` is leader misbehavior, and can/should also result in `try_skip_window()` the other three could have been created by whoever sent them to us (or in the case of `Duplicate` are perfectly innocuous because it might even just be because we already reconstructed the shred), if that's on an authenticated connection it could be used to blacklist the validator, so a distinction like `LeaderError`/`ByzantineShred` and `RelayError`/`GarbageShred` (equivalent to the original `InvalidShred` vs `InvalidSignature`), ... is something we probably want

src/consensus/blockstore.rs
qkniep: we don't want `warn` level logs for all of the others, specifically not for `Duplicate`, I was already considering to demote it from `debug` to `trace`

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -58,6 +32,18 @@ pub struct SlotBlockData {
     pub(super) equivocated: bool,
 }
 
+#[derive(Debug)]
+pub(super) enum AddShredResult {
+    FirstShred(Slot),
```
qkniep: maybe avoid this by passing slot number "manually" to `handle_add_shred_result()`?

src/consensus/blockstore.rs
qkniep: what we probably want to do instead, both here and for `LeaderError` (or whatever we will call it), is to ensure that `try_skip_window()` is called

the easiest way, to do this is to just mark the block (for equivocation we already have the `equivocated` flag in blockstore) and then no longer ingest shreds, this ensures that if we see such an error before reconstructing the block, we will not reconstruct it and after the timeout (but only then) initiate `try_skip_window()`

the more solid way (circumventing the timeout and also triggering if we already voted notar on this specific block, but maybe not the rest of the window) would be to directly initiate `try_skip_window()` from here, for this we probably need a new `VotorEvent`, something like `LeaderMisbehavior`

PR #221, opened 2025-09-25
refactor: do not construct potentially invalid `Shred` in `reed_solomon_shred()`

akhi3030:
`reed_solomon_shred()` is returning `DataShred` and `CodingShred` and assigning `index_in_shred` in them.  It does not have enough context to assign the right index in the shreds.  The actual index depends on the actual type of shredder that we are using.

See https://github.com/qkniep/alpenglow/pull/190/files#r2376067497 for more context.

This PR updates `reed_solomon_shred()` to return just the shredded payload instead.  Then the individual shredders construct the actual `Shred`s with valid indexes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aes"
-version = "0.9.0-rc.0"
+version = "0.9.0-rc.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd4838e4ad37bb032dea137f441d5f71c16c26c068af512e64c5bc13a88cdfc7"
+checksum = "7e713c57c2a2b19159e7be83b9194600d7e8eb3b7c2cd67e671adf47ce189a05"
 dependencies = [
  "cfg-if",
  "cipher",
@@ -83,7 +83,7 @@ dependencies = [
  "sha2 0.10.9",
  "smallvec",
  "static_assertions",
- "thiserror 2.0.12",
+ "thiserror 2.0.16",
  "time",
  "tokio",
  "tokio-util",
@@ -92,9 +92,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.19"
+version = "0.6.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "301af1932e46185686725e0fad2f8f2aa7da69dd70bf6ecc44d6b703844a3933"
+checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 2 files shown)


src/shredder/reed_solomon.rs
```diff
@@ -28,14 +28,21 @@ pub(super) enum ReedSolomonDeshredError {
     TooMuchData,
 }
 
+/// The data and coding shreds returned from [`reed_solomon_shred()`] on success.
+pub(super) struct Shreds {
```
akhi3030: not a big fan of this name.  If you can come up with something better, that will be great.

Also I had to introduce this because of clippy.  Without it, clippy complained that the signature for `reed_solomon_shred()` is too complex.

qkniep: I actually agree with clippy, it's good to have a type for this. I would probably just call it `ReedSolomonShreds` or `ReedSolomonOutput`?!

src/shredder/reed_solomon.rs
```diff
@@ -48,26 +55,10 @@ pub(super) fn reed_solomon_shred(
...
-    let coding_shreds = coding_parts
-        .into_iter()
-        .enumerate()
-        .map(|(i, p)| CodingShred(payload_from_index_and_data(i, p)))
-        .collect();
-
-    Ok((data_shreds, coding_shreds))
+    let coding = rs::encode(num_data, num_coding, payload.chunks(shred_bytes)).unwrap();
+
+    let data = payload.chunks(shred_bytes).map(|c| c.to_vec()).collect();
```
qkniep: maybe cleaner?
```rust
        let data = payload.chunks(shred_bytes).map(|c| c.to_vec()).collect();
        let coding = rs::encode(num_data, num_coding, &data).unwrap();
```

src/shredder/reed_solomon.rs
qkniep: if we want something shorter, `RsShreds` as you used for the import might work, or maybe `RawShreds`...

src/shredder.rs
```diff
@@ -505,47 +499,55 @@ fn data_and_coding_to_output_shreds(
...
+fn build_merkle_tree(rs_shreds: &RsShreds) -> MerkleTree {
     // zero-allocation chaining of slices
-    let leaves = data_shreds
+    let leaves = rs_shreds
+        .data
         .iter()
-        .map(|d| d.0.data.as_ref())
-        .chain(coding_shreds.iter().map(|c| c.0.data.as_ref()));
+        .map(|d| d.as_ref())
+        .chain(rs_shreds.coding.iter().map(|c| c.as_ref()));
```
qkniep: maybe this is slightly more readable?
```rust
    let leaves = rs_shreds
        .data
        .iter()
        .chain(&rs_shreds.coding)
        .map(Vec::as_slice);

```

qkniep (approved):
LGTM, I really like the new API! Hated the leaking of abstractions before and the `DataShred` and `CodingShred` types, do we still use those anywhere or can we remove them?

I feel like there might be a way to make `data_and_coding_to_output_shreds` and `create_output_shreds_for_other_leader` more concise, but right now I can't come up with something better.

src/shredder/reed_solomon.rs
akhi3030: `RawShreds` FTW.

PR #231, opened 2025-09-29
feat: stop accepting disseminated shreds for a misbehaving leader

akhi3030:
previously we were discarding shreds from a leader that showed equivocation.  Now, we also handle leader that misbehaved.

Part of #85

### DIFF
--- a/src/consensus/blockstore/slot_block_data.rs
+++ b/src/consensus/blockstore/slot_block_data.rs
@@ -80,8 +80,8 @@ impl SlotBlockData {
     ) -> Result<Option<VotorEvent>, AddShredError> {
         assert_eq!(shred.payload().header.slot, self.slot);
         if self.leader_misbehaved {
-            debug!("recevied shred from equivocating leader, not adding to blockstore");
-            return Err(AddShredError::Equivocation);
+            debug!("recevied shred from misbehaving leader, not adding to blockstore");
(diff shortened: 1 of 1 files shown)


src/consensus/blockstore/slot_block_data.rs
```diff
@@ -78,16 +79,17 @@ impl SlotBlockData {
         leader_pk: PublicKey,
     ) -> Result<Option<VotorEvent>, AddShredError> {
         assert_eq!(shred.payload().header.slot, self.slot);
-        if self.equivocated {
+        if self.leader_misbehaved {
             debug!("recevied shred from equivocating leader, not adding to blockstore");
```
qkniep: this log needs updating

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -78,16 +79,17 @@ impl SlotBlockData {
...
             return Err(AddShredError::Equivocation);
```
qkniep: maybe we should change the error returned here, not 100% sure to what, but `AddShredError::InvalidShred` would already be more accurate

akhi3030: I don't want to introduce a new variant for this case.  Switching to `InvalidShred` makes sense.

PR #232, opened 2025-09-29
feat: introduce a new type `ValidatedShreds`

### LINKED ISSUE #225: Add tests for `DeshredError::InvalidLayout`
We introduced a new error condition where if the shreds array is not in the expected layout then deshreding fails.  We should add tests to make sure that the validation works properly.

### LINKED ISSUE #226: Consider adding a new type for validated shreds array
`deshred_validated_shreds()` should be called with a new type e.g. `ValidatedShreds`.  This type can only be constructed by validating the layout of the shreds array.

akhi3030:
Closes #226
Closes #225

### DIFF
--- a/src/shredder.rs
+++ b/src/shredder.rs
@@ -251,8 +251,7 @@ impl Shredder for RegularShredder {
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
-        let raw_shreds =
-            reed_solomon_shred(payload.into(), DATA_SHREDS, TOTAL_SHREDS - DATA_SHREDS)?;
+        let raw_shreds = reed_solomon_shred(payload.into(), Self::CODING_OUTPUT_SHREDS)?;
         Ok(data_and_coding_to_output_shreds(header, raw_shreds, sk))
     }
 
@@ -260,8 +259,7 @@ impl Shredder for RegularShredder {
         shreds: ValidatedShreds,
     ) -> Result<(Slice, [ValidatedShred; TOTAL_SHREDS]), DeshredError> {
         let shreds = shreds.to_shreds();
-        let payload =
-            reed_solomon_deshred(shreds, DATA_SHREDS, TOTAL_SHREDS - DATA_SHREDS, DATA_SHREDS)?;
+        let payload = reed_solomon_deshred(shreds, Self::CODING_OUTPUT_SHREDS)?;
 
         // deshreding succeeded above, there should be at least one shred in the array so the unwrap() below should be safe
         let any_shred = shreds.iter().find_map(|s| s.as_ref()).unwrap();
@@ -270,8 +268,7 @@ impl Shredder for RegularShredder {
         // additional Merkle tree validity check
         let merkle_root = any_shred.merkle_root;
         let (header, payload) = slice.clone().deconstruct();
-        let raw_shreds =
-            reed_solomon_shred(payload.into(), DATA_SHREDS, TOTAL_SHREDS - DATA_SHREDS)?;
+        let raw_shreds = reed_solomon_shred(payload.into(), Self::CODING_OUTPUT_SHREDS)?;
         let tree = build_merkle_tree(&raw_shreds);
         if tree.get_root() != merkle_root {
             return Err(DeshredError::InvalidMerkleTree);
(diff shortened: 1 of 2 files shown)


src/shredder.rs
```diff
@@ -218,34 +227,16 @@ pub trait Shredder {
...
-            }
-        }
-        for shred in shreds.iter().take(Self::DATA_OUTPUT_SHREDS) {
-            if let Some(shred) = shred
-                && !matches!(&shred.payload_type, ShredPayloadType::Data(_p))
-            {
-                return Err(DeshredError::InvalidLayout);
-            }
-        }
-        for shred in shreds.iter().skip(Self::CODING_OUTPUT_SHREDS) {
```
akhi3030: this was a bug in the original code.  This PR helped catch it 🎉

src/shredder/validated_shreds.rs
```diff
@@ -0,0 +1,89 @@
...
+                && !matches!(&shred.payload_type, ShredPayloadType::Coding(_p))
+            {
+                return None;
+            }
+        }
+        Some(Self(shreds))
+    }
+
+    /// Returns the inner array of [`ValidatedShred`]s.
+    pub(super) fn into_shreds(self) -> &'a [Option<ValidatedShred>; TOTAL_SHREDS] {
```
qkniep: I would expect `into_*` to return an owned value, maybe `as_shreds()` instead? then taking `&self` might also be more intuitive (even though it's probably one more copy of a pointer)

src/shredder.rs
```diff
@@ -150,6 +153,12 @@ impl Shred {
             ShredPayloadType::Coding(p) | ShredPayloadType::Data(p) => p,
         }
     }
+
+    pub const fn payload_mut(&mut self) -> &mut ShredPayload {
```
qkniep, line 147:
```suggestion
    /// Mutably references the payload contained in this shred.
    pub const fn payload_mut(&mut self) -> &mut ShredPayload {
```

src/shredder.rs
qkniep, line 140:
```suggestion
    /// References the payload contained in this shred.
    pub const fn payload(&self) -> &ShredPayload {
```

src/shredder/validated_shreds.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl<'a> ValidatedShreds<'a> {
+    /// Creates a new [`ValidatedShreds`].
+    pub(super) fn try_new(
+        shreds: &'a [Option<ValidatedShred>; TOTAL_SHREDS],
+        data_shreds: usize,
+        coding_shreds: usize,
+    ) -> Option<Self> {
+        if data_shreds + coding_shreds != TOTAL_SHREDS {
+            return None;
```
qkniep: should probably just panic here, this is 100% an error in the client implementation, not in the shreds

src/shredder/validated_shreds.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        for (i, shred) in shreds.iter().enumerate() {
+            if let Some(shred) = shred
+                && *shred.payload().shred_index != i
+            {
+                return None;
```
qkniep: This is also a bit problematic, as this can never be the result of a wrong leader, just means that we made a mistake placing shreds in the wrong places (however, upstream we treat `InvalidLayout` as `InvalidShred` and thus leader misbehavior). The options I see are:
- Try to reorder them (requires mutable access to the `shreds` slice, still need to handle failure if we have duplicate shreds not just wrong order).
- Panic instead of returning `None`.
- Return a `Result` with at least 2 different errors here, instead of an `Option`.

src/shredder/validated_shreds.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        // the shred index is invalid
+        let mut shreds = RegularShredder::shred(slice.clone(), &sk).unwrap();
+        shreds[1].payload_mut().shred_index = ShredIndex::new(0).unwrap();
+        let shreds = shreds.map(Some);
+        assert!(
+            ValidatedShreds::try_new(&shreds, DATA_SHREDS, TOTAL_SHREDS - DATA_SHREDS).is_none()
+        );
+
+        // there are more data shreds in the array
```
qkniep:
```suggestion
        // there are data shreds in coding shred positions in the array
```

src/shredder/validated_shreds.rs
```diff
@@ -0,0 +1,89 @@
...
+        let shreds = RegularShredder::shred(slice.clone(), &sk)
+            .unwrap()
+            .map(Some);
+        assert!(ValidatedShreds::try_new(&shreds, 1, TOTAL_SHREDS - 1).is_none());
+
+        // there are fewer data shreds in the array
```
qkniep:
```suggestion
        // there are coding shreds in data shred locations in the array
```

src/shredder/validated_shreds.rs
akhi3030: Good point.  I think panicking is the best.  If the shreds are not in the right place, then there is a bug in the node's code.  And trying to reorder things is just trying to cope with buggy code which is probably not going to be helpful.

src/shredder/validated_shreds.rs
akhi3030: picked `to_shreds()` a style we are using elsewhere too.

PR #236, opened 2025-09-29
nit: make `ValidatedShreds` constructer simpler

qkniep:
Is this better or worse?


akhi3030:
I like it, much cleaner.

PR #238, opened 2025-10-01
nit: make derive order more consistent

qkniep:


akhi3030 (approved):
what is the preference for the ordering 😄

PR #245, opened 2025-10-08
refactor: generalize Monte-Carlo quorum robustness testing

qkniep:
Generalizes the code written specifically to test Rotor robustness against liveness and equivocation attacks. Instead, we now have a general `struct QuorumRobustnessTest`, which can be used to test robustness of various quorum-based protocols against an adversary corrupting a certain subset of some of these quorums.

What is not yet supported is to sample different quorums within one protocol according to different sampling strategies. Right now, different quorums can have their own instances of type `S: SamplingStrategy`, however, the type is fixed for the entire `QuorumRobustnessTest`.

### DIFF
--- a/src/bin/simulations/quorum_robustness.rs
+++ b/src/bin/simulations/quorum_robustness.rs
@@ -44,7 +44,7 @@ pub struct AdversaryStrength {
 
 /// Test harness for quorum robustness testing.
 pub struct QuorumRobustnessTest<S: SamplingStrategy> {
-    samplers: RwLock<Vec<S>>,
+    samplers: Vec<S>,
     quorum_samplers: Vec<usize>,
     quorum_sizes: Vec<usize>,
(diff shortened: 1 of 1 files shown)


src/bin/simulations/quorum_robustness.rs
```diff
@@ -0,0 +1,533 @@
...
+pub struct AdversaryStrength {
+    /// Fraction of stake that may crash.
+    pub crashed: f64,
+    /// Fraction of stake that may be arbitrarily controlled by the adversary.
+    pub byzantine: f64,
+}
+
+/// Test harness for quorum robustness testing.
+pub struct QuorumRobustnessTest<S: SamplingStrategy> {
+    samplers: RwLock<Vec<S>>,
```
akhi3030: nit:doesn't look like this needs a lock.  But that might change in the future?

PR #246, opened 2025-10-12
replace `bincode` with `wincode`

qkniep:
The Anza-developed crate `wincode` promises a big performance boost while mostly being a drop-in replacement for `bincode`, especially keeping 100% compatibility with the wire format of `bincode`.

My first benchmark shows more than 10x speedup for both serialization and deserialization of a dummy type roughly matching an Alpenglow certificate.
Screenshot 2025-10-12 at 16 41 25
Screenshot 2025-10-12 at 16 41 43

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3266,7 +3266,7 @@ dependencies = [
 [[package]]
 name = "wincode"
 version = "0.1.2"
-source = "git+https://github.com/anza-xyz/wincode?branch=master#3e6e6850059fca823db983dc74f60d22ab2839a7"
+source = "git+https://github.com/anza-xyz/wincode?rev=3e6e6850059fca823db983dc74f60d22ab2839a7#3e6e6850059fca823db983dc74f60d22ab2839a7"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -3277,7 +3277,7 @@ dependencies = [
 [[package]]
 name = "wincode-derive"
 version = "0.1.1"
-source = "git+https://github.com/anza-xyz/wincode?branch=master#3e6e6850059fca823db983dc74f60d22ab2839a7"
+source = "git+https://github.com/anza-xyz/wincode?rev=3e6e6850059fca823db983dc74f60d22ab2839a7#3e6e6850059fca823db983dc74f60d22ab2839a7"
 dependencies = [
  "darling",
  "proc-macro2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ time = { version = "0.3", features = ["serde"] }
 tokio = { version = "1", features = ["full"] }
 tokio-util = "0.7"
 toml = "0.9"
-wincode = { git = "https://github.com/anza-xyz/wincode", branch = "master", features = ["derive"] }
+wincode = { git = "https://github.com/anza-xyz/wincode", rev = "3e6e6850059fca823db983dc74f60d22ab2839a7", features = ["derive"] }
 
 [dev-dependencies]
 divan = "0.1"


qkniep:
My PR has landed, only waiting for next release of `wincode` now.

benches/network.rs
```diff
@@ -1,51 +1,86 @@
...
-            rng.fill_bytes(&mut hash);
-            let sk = aggsig::SecretKey::new(&mut rng);
-            let vote = Vote::new_notar(Slot::new(0), hash.into(), &sk, 0);
-            ConsensusMessage::Vote(vote)
+            let msg = ConsensusMessage::Vote(generate_vote());
+            wincode::serialize(&msg).unwrap()
         })
-        .bench_values(|msg: ConsensusMessage| bincode::serde::encode_to_vec(msg, BINCODE_CONFIG));
+        .bench_values(|bytes: Vec<u8>| {
+            let _msg: ConsensusMessage = wincode::deserialize(&bytes).unwrap();
```
akhi3030: does this deserialize the same vote again and again?  Are we worried about some weird caching effects giving a false impression that this is faster than it really is.

src/consensus/block_producer.rs
```diff
@@ -392,17 +390,12 @@ where
...
+    let parent_encoded_len = wincode::serialize(&parent).unwrap().len();
 
-    // HACK: As long as the size of the txs vec fits in a single byte,
-    // bincode encoding seems to take a single byte so account for that here.
-    assert_eq!(highest_non_zero_byte(MAX_TRANSACTIONS_PER_SLICE), 1);
     let mut slice_capacity_left = MAX_DATA_PER_SLICE
         .checked_sub(parent_encoded_len)
         .unwrap()
-        .checked_sub(1)
+        .checked_sub(8)
```
akhi3030: some documentation please.

src/consensus/block_producer.rs
```diff
@@ -535,16 +526,16 @@ mod tests {
             produce_slice_payload(&txs_receiver, parent.clone(), duration_left).await;
         assert_eq!(maybe_duration, Duration::ZERO);
         assert_eq!(payload.parent, parent);
-        // bin encoding an empty Vec takes 1 byte
-        assert_eq!(payload.data.len(), 1);
+        // bin encoding an empty Vec takes 8 bytes
+        assert_eq!(payload.data.len(), 8);
```
akhi3030, line 534: I wonder how stable the schema for wincode will be moving forward.  If we imagine that it can change then maybe we should add some tests that will catch potential issues relating to how many bytes it takes to encode an object.  We can do it in separate PRs of course.  Or even better, if wincode could actually expose an API that can report how many bytes an object will take without encoding it.

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
 //
 // Deriving PartialEq and Eq to support testing.
 // It only makes sense beccause the underlying signature scheme happens to be deterministic and unique.
-// Revaluate if we change the signature scheme.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+// Reevaluate if we change the signature scheme.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct IndividualSignature(pub BlstSignature);
```
akhi3030: not related to the PR but this also type checks maybe we should make the change?


```suggestion
pub struct IndividualSignature(BlstSignature);
```

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
 
+impl<'de> SchemaRead<'de> for IndividualSignature {
+    type Dst = IndividualSignature;
+
+    fn read(
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        let sig_bytes = reader.read_borrowed(96)?;
```
akhi3030: docs please on the 96

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
+        let sig = BlstSignature::deserialize(sig_bytes)
+            .map_err(|_| wincode::ReadError::Custom("invalid BLS encoding"))?;
```
akhi3030: It seems like deserialize can return different types of errors.  But `Custom` can only take a `'static str`, maybe we can add a log statement here to capture the error?

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
+        dst.write(IndividualSignature(sig));
+        wincode::ReadResult::Ok(())
+    }
+}
+
+impl SchemaWrite for IndividualSignature {
+    type Src = IndividualSignature;
+
+    fn size_of(_src: &Self::Src) -> wincode::WriteResult<usize> {
+        Ok(96)
```
akhi3030: define a constant for 96 to use here and in `fn read()` above?

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
+    }
+
+    fn write(writer: &mut wincode::io::Writer, src: &Self::Src) -> wincode::WriteResult<()> {
+        unsafe { Ok(writer.write_t(&src.0.serialize())?) }
```
akhi3030: Should we consider using the [safety comment](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) here?

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        let sig_bytes = reader.read_borrowed(96)?;
+        let num_bits = <usize>::get(reader)?;
+        let bitmask_raw_vec = <Vec<usize>>::get(reader)?;
+        let sig = BlstSignature::from_bytes(sig_bytes)
+            .map_err(|_| wincode::ReadError::Custom("invalid BLS encoding"))?;
+        let mut bitmask = BitVec::try_from_vec(bitmask_raw_vec)
+            .map_err(|_| wincode::ReadError::Custom("invalid bitmap"))?;
```
akhi3030: maybe log errors like above.

src/crypto/aggsig.rs
```diff
@@ -78,19 +81,90 @@ impl PublicKey {
...
+        wincode::ReadResult::Ok(())
+    }
+}
+
+impl SchemaWrite for AggregateSignature {
+    type Src = AggregateSignature;
+
+    fn size_of(src: &Self::Src) -> wincode::WriteResult<usize> {
+        let data = src.bitmask.as_bitslice().domain().collect::<Vec<usize>>();
+        Ok(96 + 8 + <Vec<usize> as wincode::SchemaWrite>::size_of(&data)?)
```
akhi3030: docs

src/crypto/signature.rs
```diff
@@ -26,9 +29,38 @@ pub struct PublicKey(VerificationKey);
...
+
+    fn read(
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        let mut sig_bytes: MaybeUninit<[u8; 64]> = MaybeUninit::uninit();
+        let sig = unsafe {
+            reader.read_t(&mut sig_bytes)?;
+            ed25519_consensus::Signature::from(sig_bytes.assume_init())
+        };
```
akhi3030: safety comment?

src/crypto/signature.rs
```diff
@@ -26,9 +29,38 @@ pub struct PublicKey(VerificationKey);
...
+        dst.write(Signature(sig));
+        wincode::ReadResult::Ok(())
+    }
+}
+
+impl SchemaWrite for Signature {
+    type Src = Signature;
+
+    fn size_of(_src: &Self::Src) -> wincode::WriteResult<usize> {
+        Ok(64)
```
akhi3030: docs

src/crypto/signature.rs
```diff
@@ -26,9 +29,38 @@ pub struct PublicKey(VerificationKey);
...
+    }
+
+    fn write(writer: &mut wincode::io::Writer, src: &Self::Src) -> wincode::WriteResult<()> {
+        unsafe { Ok(writer.write_t(&src.0.to_bytes())?) }
```
akhi3030: safety comments

src/shredder/shred_index.rs
```diff
@@ -75,12 +77,31 @@ impl<'de> Visitor<'de> for ShredIndexVisitor {
...
+
+    fn read(
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        unsafe {
+            reader.read_t(dst)?;
+            if dst.assume_init_ref().0 >= TOTAL_SHREDS {
+                // FIXME: replace this arbitrary error type
+                Err(wincode::ReadError::InvalidCharLead(0))
```
akhi3030: do you mean the custom error type?  should be possible to do now, right?

src/shredder/shred_index.rs
```diff
@@ -75,12 +77,31 @@ impl<'de> Visitor<'de> for ShredIndexVisitor {
...
 }
 
+impl<'de> SchemaRead<'de> for ShredIndex {
+    type Dst = Self;
+
+    fn read(
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        unsafe {
```
akhi3030, line 89: safety comment?

src/types/slice.rs
```diff
@@ -157,33 +155,29 @@ impl From<&[u8]> for SlicePayload {
...
-    let highest_byte = highest_non_zero_byte(desired_size);
-    let size = if highest_byte == 1 {
-        left.checked_sub(highest_byte).unwrap()
-    } else {
-        left.checked_sub(highest_byte)
-            .unwrap()
-            .checked_sub(1)
-            .unwrap()
-    };
+    let size = left.checked_sub(8).unwrap();
```
akhi3030: docs

src/types/slice.rs
```diff
@@ -157,33 +155,29 @@ impl From<&[u8]> for SlicePayload {
...
     rng.fill_bytes(&mut data);
-    bincode::serde::encode_into_slice(data, &mut payload[used..], BINCODE_CONFIG).unwrap();
-
+    used += wincode::serialize_into(&data, &mut payload[used..]).unwrap();
+    assert_eq!(used, desired_size);
+
+    let len = payload.len();
+    let cap = payload.capacity();
+    let ptr = payload.as_mut_ptr() as *mut u8;
+    std::mem::forget(payload); // prevent dropping uninitialized memory
```
akhi3030:
```suggestion
    // prevent dropping uninitialized memory
    std::mem::forget(payload); 
```

it feels more rusty.

src/types/slice.rs
```diff
@@ -157,33 +155,29 @@ impl From<&[u8]> for SlicePayload {
...
+    let payload: Vec<u8> = unsafe { Vec::from_raw_parts(ptr, len, cap) };
```
akhi3030: docs and safety comment

akhi3030 (commented):
looks great, i left a few comments so not approving yet.  I would like to review once more.

src/shredder/shred_index.rs
qkniep: oh, I missed that one, should've made another pass

src/crypto/aggsig.rs
qkniep: Created #277 for this

src/crypto/aggsig.rs
qkniep: here I could actually get rid of the use of `unsafe`

src/crypto/signature.rs
qkniep: got rid of `unsafe` here

src/types/slice.rs
qkniep: this entire function was kind of stupid, we were essentially manually serializing a `SlicePayload` just to deserialize it in the end...

qkniep: lol, this comment got me flagged by some researchers from Stockholm university doing automatic code-of-conduct analysis on public GitHub comments...

src/consensus/block_producer.rs
qkniep: Since `wincode` was specifically built to be compatible with the specific `bincode` configuration that was used by agave, I would expect it to not change in the near/medium-term future.

src/types/slice.rs
akhi3030: WTH?!?!

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -307,7 +306,7 @@ impl BlockData {
             .values()
             .map(|s| s.merkle_root.as_ref().unwrap());
         let tree = DoubleMerkleTree::new(merkle_roots);
-        let block_hash = tree.get_root();
+        let block_hash = tree.get_root().clone();
```
```suggestion
        let block_hash = tree.get_root();
```

doesn't seem to be needed.

src/crypto/aggsig.rs
```diff
@@ -76,21 +96,122 @@ impl PublicKey {
...
+                bitmask_raw_vec.len() * usize::BITS as usize,
+            );
+            return Err(wincode::ReadError::Custom("want to use too many bits"));
+        }
+        let mut bitmask =
+            BitVec::try_from_vec(bitmask_raw_vec).expect("bitmask vector should never be too big");
+        // SAFETY: We just built `bitmask` from a fully allocated `Vec`.
+        // Also, we just checked that `num_bits` is not too big.
+        unsafe {
+            bitmask.set_len(num_bits);
```
akhi3030: I am wondering if `num_bits` is guaranteed to be smaller than the current length and then we could use `truncate` which is `safe` and should be just as performant.

src/crypto/aggsig.rs
```diff
@@ -76,21 +96,122 @@ impl PublicKey {
...
+impl<'de> SchemaRead<'de> for AggregateSignature {
+    type Dst = AggregateSignature;
+
+    fn read(
+        reader: &mut wincode::io::Reader<'de>,
+        dst: &mut MaybeUninit<Self::Dst>,
+    ) -> wincode::ReadResult<()> {
+        // read raw data
+        let sig_bytes = reader.read_borrowed(UNCOMPRESSED_SIG_SIZE)?;
+        let num_bits = <usize>::get(reader)?;
```
akhi3030, line 152: nit: I wonder if we should consider separating out serializing and deserilizing a `BitVec` and then use those functions here.  This is probably the only user but it might be a bit cleaner.  We can also do it in subsequent PRs or just create a github issue.

src/types/slot.rs
```diff
@@ -17,7 +17,8 @@ pub const SLOTS_PER_WINDOW: u64 = 4;
 pub const SLOTS_PER_EPOCH: u64 = 18_000;
 
 /// Slot number type.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, SchemaRead, SchemaWrite)]
+#[repr(transparent)]
 pub struct Slot(u64);
```
akhi3030, line 22: does wincode handle u64 natively and not usize?  Is that why you didn't have to implement custom impl of SchemaRead here?  If that is the case, I wonder if we can change the types of `SliceIndex` to u64 and provide some wrappers to convert to usize to minimise the amount of code we have to maintain.  As we are assuming we will only ever support 64 bit architectures.  We can also do it in future PRs.

src/shredder.rs
```diff
@@ -53,7 +53,7 @@ pub const MAX_DATA_PER_SHRED: usize = 1024;
 /// Maximum number of bytes an entire slice can hold, incl. padding.
 pub const MAX_DATA_PER_SLICE_AFTER_PADDING: usize = DATA_SHREDS * MAX_DATA_PER_SHRED;
 /// Maximum number of payload bytes a slice can hold.
-pub const MAX_DATA_PER_SLICE: usize = MAX_DATA_PER_SLICE_AFTER_PADDING - 1;
+pub const MAX_DATA_PER_SLICE: usize = MAX_DATA_PER_SLICE_AFTER_PADDING - 8;
```
akhi3030: nit: document the 8 please.

src/types/slot.rs
qkniep: no, the reason is that `Slot` does not have an upper bound (any valid `u64` is also a valid `Slot`), unlike `SliceIndex`, we only need the custom impl to check the upper bound

akhi3030: ah yes, I forgot about that, makes sense.

src/crypto/aggsig.rs
qkniep: the problem is, truncate would only work in multiples of `usize::BITS`, but `num_bits` can be anything in-between... that also the only good reason I can see why `BitVec` uses the separate `num_bits` at all

akhi3030: got it.  Maybe we should document that?

src/shredder.rs
qkniep: good catch! I actually misinterpreted the 1 myself, it's the padding byte not a length encoding...

documented this now and also fixed a bug in `block_producer`, where we were not accounting for encoding the tx length

src/crypto/aggsig.rs
qkniep: created #280

PR #252, opened 2025-10-16
refactor: type-safe Merkle trees

qkniep:
Makes the `MerkleTree` type generic over types implementing a `trait MerkleTreeType`. This allows users to create their own type-safe instantiations of the `MerkleTree` type with specific types for leaf nodes, root, and proofs, instead of just using `Hash` and `&[u8]` everywhere.

Next steps in this PR would be:
- [x] replace remaining occurences of `Hash`, `Vec<Hash>`, and `&[u8]` in the public API of `MerkleTree`
- [ ] ~replace `type Root` in`SliceMerkleTreeType` with `SliceRoot` newtype (#248)~
- [ ] ~replace `type Root` in `DoubleMerkleTreeType` with `BlockHash` newtype (#249)~
- [ ] ~replace `type Proof` in both `SliceMerkleTreeType` and `DoubleMerkleTreeType` with newtypes (#250)~

Remaining points are instead addressed in a follow-up PR, #253.

### DIFF
--- a/src/crypto/merkle.rs
+++ b/src/crypto/merkle.rs
@@ -100,15 +100,21 @@ const EMPTY_ROOTS: [Hash; MAX_MERKLE_TREE_HEIGHT] = [
 /// Marker trait for the leaf nodes of a Merkle tree.
 pub trait MerkleLeaf: AsRef<[u8]> {}
 
-/// Marker trait for the root of a Merkle tree.
-pub trait MerkleRoot: From<Hash> + Into<Hash> + Clone {}
+/// Trait for the root of a Merkle tree.
+pub trait MerkleRoot: From<Hash> {
+    fn as_hash(&self) -> &Hash;
+}
 
 /// Marker trait for the proof of a Merkle tree.
 pub trait MerkleProof: AsRef<[Hash]> + From<Vec<Hash>> {}
 
 impl MerkleLeaf for Vec<u8> {}
 impl MerkleLeaf for Hash {}
-impl MerkleRoot for Hash {}
+impl MerkleRoot for Hash {
+    fn as_hash(&self) -> &Hash {
+        self
+    }
+}
 impl MerkleProof for Vec<Hash> {}
 
 /// A plain Merkle tree over arbitrary bytes.
@@ -260,7 +266,7 @@ impl<Leaf: MerkleLeaf, Root: MerkleRoot, Proof: MerkleProof> MerkleTree<Leaf, Ro
             };
             i /= 2;
         }
-        node == root.clone().into()
+        node == *root.as_hash()
     }
 
     /// Checks a Merkle path proves the given leaf's data is last in the tree.
@@ -287,7 +293,7 @@ impl<Leaf: MerkleLeaf, Root: MerkleRoot, Proof: MerkleProof> MerkleTree<Leaf, Ro
             };
             i /= 2;
         }
-        node == root.clone().into()
+        node == *root.as_hash()
     }
 
     /// Hashes some leaf data with a label into a leaf node.


src/crypto/merkle.rs
```diff
@@ -93,36 +97,64 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
     hex!("33ebfb34d3aa119cb665d564acdd77b318dc86aa6a97744edb3bc03e97d776ca"),
 ];
 
+pub trait MerkleTreeType {
```
akhi3030: gotta be honest, I don't remember seeing such a use for traits before.

It seems like we are using this trait to enforce that any struct that implements it provides the three necessary fields.  then we are able to use this trait to implement the `MerkleTree` struct which uses the fields.  

Pretty slick!  It will be helpful to have this be spelled out a bit in comments when you get to that level of the PR implementation.

akhi3030 (commented):
Looks good so far.  No particular comment really.  Probably because I am not super familiar with using traits in this manner.  I think once more of the PR is fleshed out, then I may have more insights.  

I am wondering how I would've solved this problem using more simpler rust constructs.  The ways I am thinking of would result in a lot more code duplication and less safety.  So I like this.

src/crypto/merkle.rs
```diff
@@ -93,36 +97,58 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
...
+impl MerkleRoot for Hash {}
+impl MerkleProof for Vec<Hash> {}
+
+pub type PlainMerkleTree = MerkleTree<Vec<u8>, Hash, Vec<Hash>>;
+pub type SliceMerkleTree = MerkleTree<Vec<u8>, Hash, Vec<Hash>>;
+pub type DoubleMerkleTree = MerkleTree<Hash, Hash, Vec<Hash>>;
+
 /// Implementation of a Merkle tree.
-pub struct MerkleTree {
+pub struct MerkleTree<L: AsRef<[u8]>, R: MerkleRoot, P: MerkleProof> {
```
akhi3030: Shouldn't this be the following?

```suggestion
pub struct MerkleTree<L: MerkleLeaf, R: MerkleRoot, P: MerkleProof> {
```

This will require a couple of other minor changes to compile.

src/crypto/merkle.rs
```diff
@@ -93,36 +97,58 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
...
+/// Marker trait for the proof of a Merkle tree.
+pub trait MerkleProof: AsRef<[Hash]> + From<Vec<Hash>> {}
+
+impl MerkleLeaf for Vec<u8> {}
+impl MerkleRoot for Hash {}
+impl MerkleProof for Vec<Hash> {}
+
+pub type PlainMerkleTree = MerkleTree<Vec<u8>, Hash, Vec<Hash>>;
+pub type SliceMerkleTree = MerkleTree<Vec<u8>, Hash, Vec<Hash>>;
+pub type DoubleMerkleTree = MerkleTree<Hash, Hash, Vec<Hash>>;
```
akhi3030, line 136: nit: some documentation.

src/crypto/merkle.rs
```diff
@@ -93,36 +97,58 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
...
-impl MerkleTree {
+impl<L: AsRef<[u8]>, R: MerkleRoot, P: MerkleProof> MerkleTree<L, R, P> {
     /// Creates a new Merkle tree from the given data for each leaf.
     ///
     /// This will always create a perfect binary tree (filling with empty leaves as necessary).
     /// If you want to create a tree with more than half of the leaves empty,
     /// you have to explicitly pass in empty leaves as part of `data`.
-    pub fn new(data: &[impl AsRef<[u8]>]) -> Self {
-        Self::new_from_iter(data.iter().map(|leaf| leaf.as_ref()))
+    pub fn new(leaves: &[L]) -> Self {
```
akhi3030: nit: seems to only be used in tests.  I wonder if it will be a lot of work to update the tests to use `new_from_iter()` instead.

src/crypto/merkle.rs
```diff
@@ -93,36 +97,58 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
...
     /// All hashes in the tree, leaf hashes and inner nodes.
     nodes: Vec<Hash>,
     /// For each level, has the offset in `nodes` and the number of hashes on that level.
     levels: Vec<(u32, u32)>,
+    /// Marker for the type of the tree.
+    _type: PhantomData<(L, R, P)>,
 }
 
-impl MerkleTree {
+impl<L: AsRef<[u8]>, R: MerkleRoot, P: MerkleProof> MerkleTree<L, R, P> {
```
akhi3030:
```suggestion
impl<Leaf: AsRef<[u8]>, Root: MerkleRoot, Proof: MerkleProof> MerkleTree<Leaf, Root, Proof> {
```

Definitely not the usual style but these types are used a lot in the functions below and having them be a bit more descriptive like this can help with readability a lot.  

Not so important though.  I think once one stares at the code for long enough, they can get used to the single letter types.

src/crypto/merkle.rs
```diff
@@ -172,7 +203,7 @@ impl MerkleTree {
     ///
     /// The proof is the Merkle path from the leaf to the root.
     #[must_use]
-    pub fn create_proof(&self, index: usize) -> Vec<Hash> {
+    pub fn create_proof(&self, index: usize) -> P {
```
akhi3030: not part of the PR.

We could use an iterator and collect instead of the for loop if we care about performance.  We could also instead do `Vec::with_capacity(self.height())` instead of `new()`.

src/crypto/merkle.rs
```diff
@@ -93,36 +97,58 @@ const EMPTY_ROOTS: [Hash; MAX_TREE_HEIGHT] = [
...
-    pub fn new_from_iter<'a>(data: impl IntoIterator<Item = &'a [u8]>) -> Self {
+    pub fn new_from_iter<'a>(data: impl IntoIterator<Item = &'a L>) -> Self
+    where
+        L: 'a,
+    {
         let mut nodes = Vec::new();
         let mut levels = Vec::new();
 
         // calculate leaf hashes
         for leaf in data {
```
akhi3030: not part of PR.

We could use `into_iter()` and collect() here to avoid potential reallocations if this is perf sensitive.

src/crypto/merkle.rs
akhi3030, line 179: nit: we can move the definition of levels after this definition and should be able to estimate its capacity and then use `Vec::with_capacity` for it.

src/crypto/merkle.rs
```diff
@@ -206,61 +237,62 @@ impl MerkleTree {
     /// to the given `hash` at the given `index` in the tree corresponding to
     /// the given `root`.
     #[must_use]
-    pub fn check_hash_proof(hash: Hash, index: usize, root: Hash, proof: &[Hash]) -> bool {
+    fn check_hash_proof(hash: Hash, index: usize, root: &R, proof: &P) -> bool {
```
akhi3030: here and above, we can take `R` by value to avoid the clone below.  Currently it is just a `Hash` which derives `Copy` so it doesn't matter so much but still, it is cleaner API to avoid a hidden a clone.

src/crypto/merkle.rs
```diff
@@ -206,61 +237,62 @@ impl MerkleTree {
...
+        node == root.clone().into()
     }
 
     /// Checks a Merkle path proves the given leaf's data is last in the tree.
     ///
     /// Returns `true` iff the Merkle proof is valid and `index` is the last leaf in the tree.
     #[must_use]
-    pub fn check_proof_last(data: &[u8], index: usize, root: Hash, proof: &[Hash]) -> bool {
-        let hash = hash_leaf(data);
+    pub fn check_proof_last(leaf: &L, index: usize, root: &R, proof: &P) -> bool {
```
akhi3030: same as above for `R`.

akhi3030 (approved):

src/crypto/merkle.rs
qkniep: of course, good catch

src/crypto/merkle.rs
qkniep: maybe requiring `Into<Hash>` on `MerkleRoot` was the wrong choice anyways, now went with an `as_hash()` method on the trait that returns `&Hash` because that's enough for checking proofs against that root

PR #255, opened 2025-10-17
chore: some clippy fixes

qkniep:


src/bin/simulations/alpenglow/bandwidth.rs
```diff
@@ -96,7 +96,7 @@ impl<L: SamplingStrategy, R: SamplingStrategy> BandwidthTest<L, R> {
     ///
     /// Writes the results to the given CSV file.
     /// This is only meaningful after `run_multiple` has been called.
-    pub fn evaluate_supported(&self, test_name: &str, csv_file: Arc<Mutex<csv::Writer<File>>>) {
+    pub fn evaluate_supported(&self, test_name: &str, csv_file: &Arc<Mutex<csv::Writer<File>>>) {
```
akhi3030, line 99:
```suggestion
    pub fn evaluate_supported(&self, test_name: &str, csv_file: &Mutex<csv::Writer<File>>) {
```

nit

src/bin/all2all_test.rs
```diff
@@ -139,7 +139,7 @@ impl Machine {
 
         // open UDP sockets
         for node in 0..NODES_PER_MACHINE {
-            let port = BASE_PORT + node as u16;
+            let port = BASE_PORT + u16::try_from(node).unwrap();
```
akhi3030, line 142: future work: a big source of flakiness in agave today is trying to bind to ports that some other test or process already took.  We should at some point try to minimise allocating ports ahead of time and then binding as much as possible.  Essentially, always doing a `bind(port=0)` and then inspecting the returned port number.

akhi3030 (approved):
curious how you found these issues.  Is it a newer clippy version?

PR #259, opened 2025-10-19
feat: change padding scheme to one that supports >=128 data shreds

qkniep:
So far, the padding scheme was including number of padding bytes in each of the last bytes, i.e. padding data to a multiple of 4 bytes might look like this:
`DD DD DD DD | DD 03 03 03`
One downside to this scheme is that the highest number of padding bytes it supports is 255. In our cases, where we want to pad to a multiple of `2 * DATA_SHREDS` this introduces the limit `DATA_SHREDS <= 127`.

The new padding scheme pads with one `0x80` byte and zero or more `0x00` bytes (this is equivalent to adding a single `1` bit and then `0` bits as necessary), for example:
`DD DD DD DD | DD 80 00 00`
This scheme works for any number of padding bytes (just as in the old scheme we always add at least one byte of padding to ensure we are always able to differentiate data from padding).

### DIFF
--- a/src/shredder/reed_solomon.rs
+++ b/src/shredder/reed_solomon.rs
@@ -12,12 +12,14 @@ use super::{
 };
 use crate::shredder::ValidatedShred;
 
+/// Errors that may be returned by [`reed_solomon_shred()`].
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Error)]
 pub(super) enum ReedSolomonShredError {
     #[error("too much data for slice")]
     TooMuchData,
 }
 
+/// Errors that may be returned by [`reed_solomon_deshred()`].
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Error)]
 pub(super) enum ReedSolomonDeshredError {
     #[error("not enough shreds to reconstruct")]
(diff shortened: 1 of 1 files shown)


src/shredder/reed_solomon.rs
```diff
@@ -45,10 +47,9 @@ pub(super) fn reed_solomon_shred(
     }
 
     // add padding
-    // TODO: use a padding scheme that can support larger slices
-    let padding_bytes = u8::try_from(2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS))
-        .expect("cannot fit number of padding bytes in u8");
-    payload.resize(payload.len() + padding_bytes as usize, padding_bytes);
+    let padding_bytes = 2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS);
+    payload.push(0x80);
+    payload.resize(payload.len() + padding_bytes.saturating_sub(1), 0);
```
akhi3030, line 65: nit: it might be worth documenting the scheme a bit here.  Essentially copying some of the PR description to help future readers.

src/shredder/reed_solomon.rs
```diff
@@ -111,7 +112,15 @@ pub(super) fn reed_solomon_deshred(
     }
 
     // remove padding
-    let padding_bytes = restored_payload[restored_payload.len() - 1] as usize;
+    let padding_bytes = restored_payload
```
akhi3030, line 131: I guess the downside of this approach is that in the worse case we have to check the last `2 * DATA_SHREDS` bytes to find the the padding.  Currently `DATA_SHREDS` is 32 so shouldn't be very expensive but could become more expensive as we increase the number.  We can worry about that later probably.

qkniep: an alternative (non-standard) scheme would be to use more than one byte (e.g. the last 2 bytes) for the number of padding bytes, everything before that could be 0 bytes or random bytes...

akhi3030: yeah, we could think about making that change in the future.  I'm happy to keep things as is in this PR.

PR #260, opened 2025-10-21
nit: store notar fallback votes in a `BTreeMap`

akhi3030:
Before we were storing them in a `Vec` and this PR changes it to a `BTreeMap`.  This should not change the performance at all as the size of the container should never get too large but helps the reader in understanding the code better and also can help with accidental bugs.

### DIFF
--- a/src/consensus/pool/slot_state.rs
+++ b/src/consensus/pool/slot_state.rs
@@ -167,11 +167,8 @@ impl SlotState {
             VoteKind::NotarFallback(_, _) => {
                 let block_hash = vote.block_hash().unwrap().clone();
                 let outputs = self.count_notar_fallback_stake(&block_hash, voter_stake);
-                assert!(
-                    self.votes.notar_fallback[v]
-                        .insert(block_hash, vote)
-                        .is_none()
-                );
+                let res = self.votes.notar_fallback[v].insert(block_hash, vote);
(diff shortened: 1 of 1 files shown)


src/consensus/pool/slot_state.rs
```diff
@@ -167,7 +167,11 @@ impl SlotState {
             VoteKind::NotarFallback(_, _) => {
                 let block_hash = vote.block_hash().unwrap().clone();
                 let outputs = self.count_notar_fallback_stake(&block_hash, voter_stake);
-                self.votes.notar_fallback[v].push((block_hash, vote));
+                assert!(
```
akhi3030: in a separate PR, I would like to combine insertion with checking for duplicates, etc.  That will be cleaner I think.

src/consensus/pool/slot_state.rs
```diff
@@ -167,7 +167,11 @@ impl SlotState {
...
+                    self.votes.notar_fallback[v]
+                        .insert(block_hash, vote)
+                        .is_none()
+                );
```
qkniep: nit: I feel like splitting this into 
```rust
let res = self.votes.notar_fallback[v].insert(block_hash, vote);
assert!(res.is_none());
```
would be slightly clearer, then we don't do important logic within the `assert!`

PR #261, opened 2025-10-21
perf: do not always re-create RS encoders/decoders

qkniep:
Part of #284.

So far, we have been implicitly creating new `ReedSolomonEncoder` and `ReedSolomonDecoder` instances every time we make a call to our `reed_solomon` module. Reusing these instances can make a big difference (~5x in some micro-benchmarks I did).

In this PR, I implement reusing these instances across the `shredder` module, including its `reed_solomon` sub-module. For this, most of the functions have been turned into methods on the respective `trait Shredder` objects and a new `ReedSolomonCoder` object type. However, benchmarking this change there seems to be no noticeable difference to before.

Looking at the flamegraph of a simple test (a tight loop continuously shredding the same slice), gives an easy answer to why this is the case: Shredding (and deshredding) of slices is currently not actually bottlenecked by the Reed-Solomon operations, with RS calls only making up ~12.5% of the runtime, whereas building the Merkle tree makes up ~50%, ed25519 signing an additional 15%, and serialization of the `SlicePayload` an additional 20%.
Screenshot 2025-10-22 at 01 13 43

### DIFF
--- a/src/shredder/reed_solomon.rs
+++ b/src/shredder/reed_solomon.rs
@@ -7,6 +7,7 @@
 //! It is mostly a wrapper around the [`reed_solomon_simd`] crate.
 
 use reed_solomon_simd::{ReedSolomonDecoder, ReedSolomonEncoder};
+use static_assertions::const_assert;
 use thiserror::Error;
 
 use super::{
@@ -58,7 +59,11 @@ impl ReedSolomonCoder {
     /// It is initialized for [`DATA_SHREDS`] data shreds and `num_coding` coding shreds.
     /// It is also initialized for up to [`MAX_DATA_PER_SHRED`] bytes per fragment.
     pub(super) fn new(num_coding: usize) -> ReedSolomonCoder {
+        // max shreds supported by RS field
+        const_assert!(DATA_SHREDS + TOTAL_SHREDS <= 65536);
+
         assert!(num_coding <= TOTAL_SHREDS);
+
         let encoder = ReedSolomonEncoder::new(DATA_SHREDS, num_coding, MAX_DATA_PER_SHRED).unwrap();
         let decoder = ReedSolomonDecoder::new(DATA_SHREDS, num_coding, MAX_DATA_PER_SHRED).unwrap();
 
@@ -91,7 +96,7 @@ impl ReedSolomonCoder {
         let shred_bytes = (payload.len() + padding_bytes).div_ceil(DATA_SHREDS);
         self.encoder
             .reset(DATA_SHREDS, self.num_coding, shred_bytes)
-            .unwrap();
+            .expect("shred size with padding should be supported");
 
         // add padding to last shreds
         let last_shreds_bytes = (2 * DATA_SHREDS).next_multiple_of(shred_bytes);
@@ -101,16 +106,23 @@ impl ReedSolomonCoder {
         last_shreds.push(0x80);
         last_shreds.resize(last_shreds_bytes, 0);
 
-        // chunk data & add coding
+        // chunk data
         let mut data = Vec::with_capacity(DATA_SHREDS);
         payload[..boundary]
             .chunks(shred_bytes)
             .chain(last_shreds.chunks(shred_bytes))
             .for_each(|chunk| {
-                self.encoder.add_original_shard(chunk).unwrap();
+                self.encoder
+                    .add_original_shard(chunk)
+                    .expect("adding correct number of chunks of currect size");
                 data.push(chunk.to_vec());
             });
-        let result = self.encoder.encode().unwrap();
+
+        // perform coding
+        let result = self
+            .encoder
+            .encode()
+            .expect("we just added enough data shreds");
         let coding = result.recovery_iter().map(<[u8]>::to_vec).collect();
 
         Ok(RawShreds { data, coding })
@@ -125,6 +137,7 @@ impl ReedSolomonCoder {
     ///
     /// If fewer than [`DATA_SHREDS`] elements in `shreds` are `Some()` then returns [`ReedSolomonDeshredError::NotEnoughShreds`].
     /// If the restored payload is larger than [`MAX_DATA_PER_SLICE_AFTER_PADDING`] then returns [`ReedSolomonDeshredError::TooMuchData`].
+    // TODO: pass ValidatedShreds here
     pub(super) fn deshred(
         &mut self,
         shreds: &[Option<ValidatedShred>; TOTAL_SHREDS],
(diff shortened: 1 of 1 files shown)


qkniep:
Would like to merge this change, even though it doesn't give the hoped-for performance improvement on its own. Created #284 to track progress on alleviating the other bottlenecks.

src/shredder/reed_solomon.rs
```diff
@@ -33,127 +33,167 @@ pub(super) enum ReedSolomonDeshredError {
...
-/// First, however, padding is added to the payload to make it a multiple of `2 * DATA_SHREDS`.
-/// Bit padding of one 1bit and as many 0 bits as needed is added.
-/// In the byte representation this looks like `[0x80, 0x00, ..., 0x00]`.
-///
-/// Errors
-///
-/// If the provided payload is larger than [`MAX_DATA_PER_SLICE_AFTER_PADDING`] then returns [`ReedSolomonDeshredError::TooMuchData`].
-pub(super) fn reed_solomon_shred(
-    payload: &[u8],
+pub struct ReedSolomonCoder {
```
akhi3030: nit: docs

akhi3030:
```suggestion
pub(super) struct ReedSolomonCoder {
```

src/shredder/reed_solomon.rs
```diff
@@ -33,127 +33,167 @@ pub(super) enum ReedSolomonDeshredError {
...
-            ShredPayloadType::Data(d) => (*d.shred_index, &d.data),
-            ShredPayloadType::Coding(_) => panic!("should be a data shred"),
-        })
-    });
+        // determine padding length & configure encoder for shred length
+        let padding_bytes = 2 * DATA_SHREDS - payload.len() % (2 * DATA_SHREDS);
+        let shred_bytes = (payload.len() + padding_bytes).div_ceil(DATA_SHREDS);
+        self.encoder
+            .reset(DATA_SHREDS, self.num_coding, shred_bytes)
+            .unwrap();
```
akhi3030: nit: maybe a comment about why unwrap is safe here.

src/shredder/reed_solomon.rs
```diff
@@ -33,127 +33,167 @@ pub(super) enum ReedSolomonDeshredError {
...
+        last_shreds.resize(last_shreds_bytes, 0);
 
-    let restored = rs::decode(DATA_SHREDS, num_coding, data.clone(), coding).unwrap();
+        // chunk data & add coding
+        let mut data = Vec::with_capacity(DATA_SHREDS);
+        payload[..boundary]
+            .chunks(shred_bytes)
+            .chain(last_shreds.chunks(shred_bytes))
+            .for_each(|chunk| {
+                self.encoder.add_original_shard(chunk).unwrap();
```
akhi3030: nit: maybe we should add comments about why the unwraps are safe when we interact with the RS crate here and in other places in this module.


I find the error handling the RS crate quite disappointing.  Everything returns the same `Error` enum and one has to look through the call tree to figure out what errors can actually happen and if they are safe to ignore.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -259,7 +259,8 @@ impl BlockData {
         // assuming caller has inserted at least one valid shred so unwrap() should be safe
         let slice_shreds = self.shreds.get_mut(&index).unwrap();
         let (reconstructed_slice, reconstructed_shreds) =
-            match RegularShredder::deshred(slice_shreds) {
+            // PERF: new shredder all the time!
```
akhi3030, line 262: Seems like storing the shredder in `BlockData` is only a few lines of code:

```
diff --git a/src/consensus/blockstore/slot_block_data.rs b/src/consensus/blockstore/slot_block_data.rs
index d0dcdaf..6498de9 100644
--- a/src/consensus/blockstore/slot_block_data.rs
+++ b/src/consensus/blockstore/slot_block_data.rs
@@ -154,6 +154,7 @@ pub struct BlockData {
     pub(super) double_merkle_tree: Option<DoubleMerkleTree>,
     /// Cache of Merkle roots for which the leader signature has been verified.
     pub(super) merkle_root_cache: BTreeMap<SliceIndex, SliceRoot>,
+    shredder: RegularShredder,
 }

 impl BlockData {
@@ -167,6 +168,7 @@ impl BlockData {
             last_slice: None,
             double_merkle_tree: None,
             merkle_root_cache: BTreeMap::new(),
+            shredder: RegularShredder::default(),
         }
     }

@@ -260,7 +262,7 @@ impl BlockData {
         let slice_shreds = self.shreds.get_mut(&index).unwrap();
         let (reconstructed_slice, reconstructed_shreds) =
             // PERF: new shredder all the time!
-            match RegularShredder::default().deshred(slice_shreds) {
+            match self.shredder.deshred(slice_shreds) {
                 Ok(output) => output,
                 Err(DeshredError::NotEnoughShreds) => return ReconstructSliceResult::NoAction,
                 rest => {
```

Can also be done in a separate PR of course.

src/consensus/blockstore/slot_block_data.rs
```diff
@@ -306,7 +307,7 @@ impl BlockData {
             .values()
             .map(|s| s.merkle_root.as_ref().unwrap());
         let tree = DoubleMerkleTree::new(merkle_roots);
-        let block_hash = tree.get_root();
+        let block_hash = tree.get_root().clone();
```
```suggestion
        let block_hash = tree.get_root();
```

doesn't seem to be needed.

src/crypto/signature.rs
```diff
@@ -10,17 +10,9 @@
 use ed25519_consensus::{SigningKey, VerificationKey};
 use rand::CryptoRng;
 use serde::{Deserialize, Serialize};
-use static_assertions::const_assert_eq;
 use wincode::containers::Pod;
 use wincode::{SchemaRead, SchemaWrite};
 
-/// Size of an ed25519 signature.
-const SIGNATURE_SIZE: usize = 64;
-const_assert_eq!(
```
akhi3030, line 19: Is this a drive by cleanup?  Doesn't seem to be related to the rest of the PR.

akhi3030 (approved):
minor nits.  Otherwise, LGTM.

src/crypto/signature.rs
qkniep: yes, found this when merging the `wincode` PR, clippy didn't flag it because `SIGNATURE_SIZE` is used in the assert

src/consensus/blockstore/slot_block_data.rs
qkniep: thanks for the suggested change, I will think about it again and make a follow-up PR

PR #263, opened 2025-10-22
refactor: do not take slice payload by value just to add padding

qkniep:
Currently, the `reed_solomon_shred()` function takes the entire slice payload by value (mutably) just to add the padding bytes into it.

Instead, we can take the payload by reference, and initialize a new buffer only for the last shreds that contain padding information (for large enough slices this should usually just be 1 shred).

### DIFF
--- a/src/types/slice.rs
+++ b/src/types/slice.rs
@@ -98,7 +98,7 @@ impl Slice {
 
 /// Struct to hold all the header payload of a [`Slice`].
 ///
-/// This is included in each [`Shred`] after shredding.
+/// This information is included in each shred after shredding.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub(crate) struct SliceHeader {
     /// Same as [`Slice::slot`].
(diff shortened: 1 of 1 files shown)


src/types/slice.rs
```diff
@@ -121,15 +136,15 @@ impl From<SlicePayload> for Vec<u8> {
     }
 }
 
-impl From<Vec<u8>> for SlicePayload {
-    fn from(payload: Vec<u8>) -> Self {
+impl From<&Vec<u8>> for SlicePayload {
```
akhi3030, line 139: should this take a slice instead?  i.e. &[u8].  It will require updating the call sites to use `.as_slice()` but I think that is more ergonomic.

src/shredder.rs
```diff
@@ -255,24 +255,25 @@ impl Shredder for RegularShredder {
 
     fn shred(slice: Slice, sk: &SecretKey) -> Result<[ValidatedShred; TOTAL_SHREDS], ShredError> {
         let (header, payload) = slice.deconstruct();
-        let raw_shreds = reed_solomon_shred(payload.into(), Self::CODING_OUTPUT_SHREDS)?;
+        let raw_shreds = reed_solomon_shred(&payload.to_bytes(), Self::CODING_OUTPUT_SHREDS)?;
```
akhi3030, line 258: we don't save much here as we are having to convert to a vec here but it seems like we have savings in when we call `reed_solomon_shred` in the `deshred` function below where we do not need to convert to a vec now.

akhi3030 (commented):
Nice!  One small nit fixup and a comment for my understanding.

PR #264, opened 2025-10-22
nit: minor clean ups in `slot_state.rs`

akhi3030:
- remove some hidden clones
- use `btree_map::Entry` to eliminate some unwraps and double lookups
- use match instead of if statement

### DIFF
--- a/src/consensus/pool/slot_state.rs
+++ b/src/consensus/pool/slot_state.rs
@@ -497,21 +497,21 @@ impl SlotState {
 
         match (skip, notar) {
             (Some(_), _) => {
-                self.sent_safe_to_notar.insert(block_hash.clone());
                 self.pending_safe_to_notar.remove(&block_hash);
+                self.sent_safe_to_notar.insert(block_hash);
                 SafeToNotarStatus::SafeToNotar
             }
             (_, Some((n, _))) => {
                 if n != &block_hash {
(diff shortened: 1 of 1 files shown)


src/consensus/pool/slot_state.rs
```diff
@@ -467,38 +468,52 @@ impl SlotState {
...
         let skip = &self.votes.skip[own_id as usize];
         let notar = &self.votes.notar[own_id as usize];
-        if skip.is_some() || notar.is_some() && &notar.as_ref().unwrap().0 != block_hash {
-            self.sent_safe_to_notar.insert(block_hash.clone());
-            self.pending_safe_to_notar.remove(block_hash);
-            SafeToNotarStatus::SafeToNotar
-        } else {
-            if skip.is_none() && notar.is_none() {
+
+        match (skip, notar) {
```
akhi3030, line 498: I am not entirely convinced if this patch is better than it was before.  We have more code duplication now but there are fewer subtleties in if statements.  I can go either way.

qkniep: I think the changes make it clearer, but it's not trivial to understand either way 😬

qkniep (approved):
LGTM

src/consensus/pool/slot_state.rs
akhi3030: I guess we will just have to leave with it 😄

PR #270, opened 2025-10-26
fix: reserve network resource after calculating start time, not before

qkniep:
When writing docs for #269, I noticed a bug in the `broadcast()` function of the `discrete_event_simulation` module. The network resource was reserved for the entire duration of the broadcast before calculating the start time for sending. This results in a wrong (delayed) start time for the broadcast.

Instead, the network resource is now correctly reserved after calculating the start time for sending. Also added a corresponding test for the `broadcast()` function.

### DIFF
--- a/src/bin/simulations/discrete_event_simulator.rs
+++ b/src/bin/simulations/discrete_event_simulator.rs
@@ -127,7 +127,7 @@ impl<P: Protocol> SimulationEngine<P> {
 
     /// Runs the simulation `iterations` times.
     ///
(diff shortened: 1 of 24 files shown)


src/bin/simulations/discrete_event_simulator.rs
akhi3030, line 420: nit: should we add documentation to explain the return type.  We might wonder in future why we are not returning an iterator.

akhi3030 (approved):
I haven't kept up with this work so unable to do a good job reviewing it.

PR #272, opened 2025-10-26
perf: enable LTO for `release` and `bench` builds

qkniep:


akhi3030 (approved):
cool, did you take any measurements?

akhi3030:
interesting... I'll try to look into this tomorrow a bit.  Interesting that it takes much longer now to to compile and CI timed out.

qkniep:
Timing out on the build is less surprising to me: LTO is expected to make builds slower and this is the first change we push in a while that forces a full rebuild of everything.

PR #273, opened 2025-10-26
docs: improve some docs + check docs in CI

qkniep:
Previously, if we had a doc string like this:
```md
/// Returns a [`Shred`].
```
and we renamed the `Shred` type to `ShredNew`, we would not see any errors.

Now we check whether docs build correctly with `#![deny(rustdoc::broken_intra_doc_links)]` enabled in CI. This check was also added to `check.sh` for local testing.


akhi3030 (approved):
very cool!

PR #285, opened 2025-10-31
refactor: pass `ValidatedShreds` with correct sizes to `ReedSolomonCoder::deshred`

qkniep:
`ValidatedShreds` could previously contain shreds of different sizes, which would never correctly deshred. Also, `ReedSolomonCoder::deshred` still took an array of `ValidatedShred`s instead of the safer `ValidatedShreds`.

Now, `ValidatedShreds` also checks that all the shreds have the same size. Further, `ReedSolomonCoder::deshred` now takes `ValidatedShreds` and can rely on all the checks it does including the size check.

Follow-up to #261.

### DIFF
--- a/src/shredder/reed_solomon.rs
+++ b/src/shredder/reed_solomon.rs
@@ -273,20 +273,22 @@ mod tests {
         assert_eq!(res.err().unwrap(), ReedSolomonShredError::TooMuchData);
     }
 
-    // #[test]
-    // fn deshred_not_enough_shreds() {
-    //     let (header, payload) = create_slice_with_invalid_txs(MAX_DATA_PER_SLICE).deconstruct();
-    //     let mut rs = ReedSolomonCoder::new(TOTAL_SHREDS - DATA_SHREDS);
-    //     let shreds = rs.shred(&payload.to_bytes()).unwrap();
-    //     let sk = SecretKey::new(&mut rand::rng());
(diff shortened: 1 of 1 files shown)


src/shredder/validated_shreds.rs
```diff
@@ -7,18 +7,36 @@ use crate::shredder::{TOTAL_SHREDS, ValidatedShred};
 
 /// Validated shreds array type.
 ///
-/// Using strong type to enforce certain constraints e.g. the shred indexes are valid.
+/// Using strong type to enforce certain constraints:
+/// - Shreds are in the correct order.
+/// - Shred indices match expected shred type.
+/// - Shreds are all the same size.
+#[derive(Clone, Copy)]
```
akhi3030, line 14: are we sure about deriving `Copy`?  This can be a pretty expensive struct to copy.  We should be explicit about it.

akhi3030 (commented):
Nice improvement!  LGTM other than the one comment about deriving `Copy`.

src/shredder/validated_shreds.rs
qkniep: `Copy` should be fine here, because it only copies the reference

akhi3030: duh... sorry, missed the references.

PR #286, opened 2025-10-31
perf: actually reuse shredder in blockstore

qkniep:
Even though #261 changed the `reed_solomon` API to allow for reuse of the shredder, blockstore is currently creating new shredder for each slice.

This change actually makes the blockstore keep only one shredder instance. So far this works, because adding shreds to the blockstore is single-threaded (always takes a `&mut` of the blockstore). It might make sense to instead have something like a `ShredderPool` holding one shredder per thread in the future.

Follow-up to #261.

### DIFF
--- a/src/shredder.rs
+++ b/src/shredder.rs
@@ -272,12 +272,11 @@ impl Shredder for RegularShredder {
         &mut self,
         shreds: ValidatedShreds,
     ) -> Result<(Slice, [ValidatedShred; TOTAL_SHREDS]), DeshredError> {
-        let shreds = shreds.to_shreds();
         let payload_bytes = self.0.deshred(shreds)?;
         let payload = SlicePayload::from(payload_bytes.as_slice());
 
         // deshreding succeeded above, there should be at least one shred in the array so the unwrap() below should be safe
-        let any_shred = shreds.iter().find_map(|s| s.as_ref()).unwrap();
+        let any_shred = shreds.to_shreds().iter().find_map(|s| s.as_ref()).unwrap();
         let slice = Slice::from_shreds(payload, any_shred);
         let header = slice.to_header();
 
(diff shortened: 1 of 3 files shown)


src/consensus/blockstore.rs
```diff
@@ -221,9 +223,10 @@ impl Blockstore for BlockstoreImpl {
     ) -> Result<Option<BlockInfo>, AddShredError> {
         let slot = shred.payload().header.slot;
         let leader_pk = self.epoch_info.leader(slot).pubkey;
+        let shredder = self.shredder.clone();
         match self
             .slot_data_mut(slot)
-            .add_shred_from_disseminator(shred, leader_pk)?
+            .add_shred_from_disseminator(shred, leader_pk, shredder)?
```
akhi3030, line 229: So it seems like we are having to hold the shredder in a lock because `slot_data_mut` is taking a mutable reference to `self` and if we pass a mutable reference to `shredder` here, then the borrow checker is not happy?

An alternative would be to `remove` the slot_block_data from the map to call `add_shred_from_disseminator` on it and then to reinsert it into the map.  

So the trade off would be between atomic operations on the lock which should not be contested vs. additional operations on the map.  I suspect atomic operations might be more expensive but that's just an intuition.  Also, with this approach, we are not leaking the lock into additional places in the code.  But it does make this function a bit more fragile as we could forget to reinsert into the map.

Below is a rough sketch of the code of the proposal. 

```
    async fn add_shred_from_disseminator(
        &mut self,
        shred: Shred,
    ) -> Result<Option<BlockInfo>, AddShredError> {
        let slot = shred.payload().header.slot;
        let leader_pk = self.epoch_info.leader(slot).pubkey;
        let mut slot_block_data = self
            .block_data
            .remove(&slot)
            .unwrap_or(SlotBlockData::new(slot));
        // note the bug below, we should not use `?` but reinsert into map before returning.
        let ret = match slot_block_data.add_shred_from_disseminator(
            shred,
            leader_pk,
            &mut self.shredder,
        )? {
            Some(event) => Ok(self.send_votor_event(event).await),
            None => Ok(None),
        };
        self.block_data.insert(slot, slot_block_data);
        ret
}
```

qkniep: I wanted to make this change forward looking (especially to support making the blockstore compatible with concurrent shred insertions in the future). While the idea of taking and re-inserting the `SlotBlockData` is interesting, it's not really compatible with concurrency. Maybe going directly with #287 is more consistent with what I was going for.

akhi3030: got it.  I reviewed #287 and I am happy to land that instead of this one.

qkniep:
Closing in favor of #287.

PR #287, opened 2025-11-02
perf: introduce `ShredderPool`

qkniep:
Even though #261 changed the `reed_solomon` API to allow for reuse of the shredder, blockstore is currently creating new shredder for each slice.

Here we actually make the blockstore keep a `ShredderPool`. So far this only has one shredder, because adding shreds to the blockstore is single-threaded anyways (always takes a `&mut` of the blockstore). In the future this could hold up to one shredder per thread used in the blockstore. Also, it might make sense to use the `ShredderPool` in other places, e.g. the block producer.

Follow-up to #261.
Supersedes #286.

### DIFF
--- a/src/consensus/blockstore.rs
+++ b/src/consensus/blockstore.rs
@@ -225,7 +225,7 @@ impl Blockstore for BlockstoreImpl {
         let leader_pk = self.epoch_info.leader(slot).pubkey;
         let mut shredder = self
             .shredders
-            .take()
+            .checkout()
             .expect("should have a shredder because of exclusive access");
         match self.slot_data_mut(slot).add_shred_from_disseminator(
             shred,
@@ -259,7 +259,7 @@ impl Blockstore for BlockstoreImpl {
         let leader_pk = self.epoch_info.leader(slot).pubkey;
         let mut shredder = self
             .shredders
-            .take()
+            .checkout()
             .expect("should have a shredder because of exclusive access");
         match self.slot_data_mut(slot).add_shred_from_repair(
             hash,
--- a/src/shredder/pool.rs
+++ b/src/shredder/pool.rs
@@ -16,7 +16,7 @@
 //!
 //! let shredder_pool = ShredderPool::<RegularShredder>::with_size(1);
 //! {
-//!     let mut shredder = shredder_pool.take().unwrap();
+//!     let mut shredder = shredder_pool.checkout().unwrap();
 //!     use_shredder(&mut (*shredder));
 //!     // shredder is automatically returned to pool when dropped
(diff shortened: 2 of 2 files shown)


src/consensus/blockstore.rs
```diff
@@ -221,10 +223,12 @@ impl Blockstore for BlockstoreImpl {
     ) -> Result<Option<BlockInfo>, AddShredError> {
         let slot = shred.payload().header.slot;
         let leader_pk = self.epoch_info.leader(slot).pubkey;
-        match self
-            .slot_data_mut(slot)
-            .add_shred_from_disseminator(shred, leader_pk)?
-        {
+        let mut shredder = self.shredders.take().unwrap();
```
akhi3030: nit: documentation on why `unwrap` is safe.

src/shredder/pool.rs
```diff
@@ -0,0 +1,133 @@
...
+            shredders: Arc::new(Mutex::new(shredders)),
+        }
+    }
+
+    /// Takes a shredder from the pool.
+    ///
+    /// The shredder is automatically returned to the pool when dropped.
+    ///
+    /// Returns [`None`] iff the pool is empty.
+    pub fn take(&self) -> Option<ShredderGuard<S>> {
```
akhi3030: `take` makes it sound like we are removing the shredder from the pool.  I initially thought that we were forgetting to return it to the pool in the code above.  I wonder if we should pick a different name.  `borrow_mut` might work better?  Not ideal either as we are not calling this function with `&mut self` which generally `borrow_mut` implies.

src/shredder/pool.rs
```diff
@@ -0,0 +1,133 @@
...
+                pool: Arc::clone(&self.shredders),
+                shredder: Some(shredder),
+            })
+    }
+}
+
+impl<S: Shredder + Default> ShredderPool<S> {
+    /// Creates a new pool with `size` shredders.
+    pub fn with_size(size: usize) -> Self {
+        let shredders = (0..size).map(|_| S::default()).collect::<Vec<_>>();
```
akhi3030:
```suggestion
        let shredders = (0..size).map(|_| S::default()).collect();
```

nit: inference seems to work here.

akhi3030 (approved):
LGTM!  Much cleaner than #286.  A couple of minor nits.

src/shredder/pool.rs
qkniep: I see your point. Maybe `checkout()` is better?

akhi3030: `checkout()` is perfect!

PR #289, opened 2025-11-26
chore: change wincode dependency back to crates.io

### LINKED ISSUE #281: Change `wincode` dependency from git to crates.io
To get #246 merged now I fixed the `wincode` dependency to a specific git checkout. As soon as the relevant changes get a crates.io release, which should happen with the next release of the `wincode` crate, we should change this back to refer to the regular crates.io releases.

qkniep:
Now that v0.2 of the `wincode` crate is released, which includes my PR (https://github.com/anza-xyz/wincode/pull/15) that we depend on, we can change the dependency back from the specific GitHub commit to a stable crates.io version.

With the v0.2 release there were also some changes to the API that we needed to adapt to.

Closes #281.

### DIFF
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -22,18 +22,18 @@ use crate::{
 };
 
 /// A simple ping network message.
-#[derive(Clone, Debug, SchemaRead, SchemaWrite)]
-pub struct Ping;
+#[derive(Clone, Debug, Default, SchemaRead, SchemaWrite)]
+pub struct Ping(pub [u8; 32]);
 
 /// A simple pong network message.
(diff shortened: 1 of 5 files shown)


src/test_utils.rs
```diff
@@ -22,18 +22,18 @@ use crate::{
 };
 
 /// A simple ping network message.
-#[derive(Clone, Debug, SchemaRead, SchemaWrite)]
-pub struct Ping;
+#[derive(Clone, Debug, Default, SchemaRead, SchemaWrite)]
+pub struct Ping(pub [u8; 32]);
```
akhi3030, line 26: why were these changes needed?  Or did you make the change to make the test more interesting?

Issue #284, opened 2025-10-30
Improve shredder performance


qkniep:
When implementing #261, with the goal of improving shredder performance, I noticed that we are currently not being bottlenecked by the Reed-Solomon (de)shredding at all. Instead the bottleneck currently seem to be the other crypto primitives that are involved, most notably hashing for the Merkle trees but also signing:

Consider the flamegraph of a simple test (a tight loop continuously shredding the same slice). Shredding (and deshredding) of slices is currently not actually bottlenecked by the Reed-Solomon operations, with RS calls only making up ~12.5% of the runtime, whereas building the Merkle tree makes up ~50%, ed25519 signing an additional 15%, and serialization of the `SlicePayload` an additional 20%.
<img width="1719" height="691" alt="Screenshot 2025-10-22 at 01 13 43" src="https://github.com/user-attachments/assets/b87f124c-8ce7-407e-b2d3-2bd010f3bfbe" />

PR #294, opened 2025-12-18
perf: reuse data shreds from deshred for Merkle check

qkniep:
Part of #284.


src/shredder/reed_solomon.rs
```diff
@@ -156,13 +157,17 @@ impl ReedSolomonCoder {
...
-                ShredPayloadType::Data(d) => (*d.shred_index, &d.data),
-                ShredPayloadType::Coding(_) => panic!("should be a data shred"),
+        // collect data shred indices+refs and feed to decoder
+        let data_refs: Vec<(usize, &[u8])> = shreds
+            .iter()
+            .take(coding_offset)
+            .filter_map(|s| {
+                s.as_ref().map(|s| match &s.payload_type {
+                    ShredPayloadType::Data(d) => (*d.shred_index, d.data.as_slice()),
+                    ShredPayloadType::Coding(_) => panic!("should be a data shred"),
```
akhi3030, line 168: nit: not part of PR but maybe we should add a comment about why the panic is safe.

akhi3030 (approved):

src/shredder/reed_solomon.rs
qkniep: added a safety comment, in a followup PR we could also move this into [`ValidatedShreds`], which might make more sense

PR #295, opened 2025-12-18
feat: remove Merkle root from shreds

### LINKED ISSUE #291: Remove `merkle_root` from `Shred`
The `merkle_root` entry in the `Shred` struct is redundant. Since it we only check whether the Merkle root reconstructed from the Merkle path matches it and the signature matches. Instead, we could directly check the signature on the reconstructed root. For any shred but the first for each slice we can then still skip signature verification by comparing the reconstructed root against the cached one, that we previously checked the signature on.

qkniep:
Closes #291.


src/types/slice.rs
```diff
@@ -58,11 +58,16 @@ impl Slice {
     }
 
     /// Creates a [`Slice`] from raw payload bytes and the metadata extracted from a shred.
+    ///
+    /// Also takes an already-computed Merkle root to avoid recomputing it.
     #[must_use]
-    pub(crate) fn from_shreds(payload: SlicePayload, any_shred: &ValidatedShred) -> Self {
+    pub(crate) fn from_shreds_with_root(
```
qkniep, line 64: we could also still have a `from_shreds()` constructor where we derive the root from `any_shred` but at all current call sites we already have the root

src/crypto/merkle.rs
```diff
@@ -367,16 +367,7 @@ impl<Leaf: MerkleLeaf, Root: MerkleRoot, Proof: MerkleProof> MerkleTree<Leaf, Ro
...
-        let mut node = hash;
-        for h in proof.as_ref() {
-            node = match i % 2 {
-                0 => Self::hash_pair(&node, h),
-                _ => Self::hash_pair(h, &node),
-            };
-            i /= 2;
-        }
-        node == *root.as_hash()
+        *Self::derive_hash_root(hash, index, proof).as_hash() == *root.as_hash()
```
akhi3030, line 370:
```suggestion
        Self::derive_hash_root(hash, index, proof).as_hash() == root.as_hash()
```

looks like a dereference is not needed.

akhi3030 (approved):
A minor nit otherwise LGTM

Issue #292, opened 2025-12-18
Change remaining type aliases to newtypes


qkniep:
The types `ValidatorId`, `Stake`, ~`Hash`, and `ShortHash`~ still use type aliases and should probably use actual `struct`s with the newtype pattern instead.

qkniep:
Closing as a result of #341 and #342.

PR #296, opened 2025-12-18
refactor: turn `Hash` and `ShortHash` into newtypes

qkniep:
Replaces the type aliases `Hash` and `ShortHash` with newtypes, as follows:
- `type Hash = [u8; 32]` becomes `struct Hash([u8; 32])`, and
- `type ShortHash = [u8; 16]` becomes `struct ShortHash([u8; 16])`.

Along the way we also introduce a `const GENESIS_BLOCK_HASH`, which was previously hardcoded as `Hash::default()` in various places.

Part of #292.

### DIFF
--- a/src/consensus/pool/parent_ready_tracker.rs
+++ b/src/consensus/pool/parent_ready_tracker.rs
@@ -175,6 +175,7 @@ impl Default for ParentReadyTracker {
 mod tests {
     use super::*;
     use crate::crypto::Hash;
+    use crate::crypto::merkle::GENESIS_BLOCK_HASH;
     use crate::types::SLOTS_PER_WINDOW;
 
     #[test]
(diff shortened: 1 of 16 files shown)


src/consensus/pool/parent_ready_tracker.rs
```diff
@@ -249,9 +250,9 @@ mod tests {
     #[test]
     fn out_of_order_notars() {
         assert_eq!(Slot::genesis().slots_in_window().count(), 4);
-        let block1 = (Slot::new(1), [1; 32].into());
-        let block2 = (Slot::new(2), [2; 32].into());
-        let block3 = (Slot::new(3), [3; 32].into());
+        let block1 = (Slot::new(1), Hash::random_for_test().into());
```
akhi3030, line 253: we have a lot of `Hash::random_for_test().into()`.  We could maybe introduce `BlockHash::random_for_test()` in a future PR.  It will be fewer characters and more readable (especially when one is looking at code outside of an IDE).

PR #297, opened 2025-12-19
ci: fuzzing

qkniep:


fuzz/fuzz_targets/consensus_message_roundtrip.rs
```diff
@@ -0,0 +1,24 @@
...
+use alpenglow::consensus::ConsensusMessage;
+use libfuzzer_sys::fuzz_target;
+
+// Fuzz serialization roundtrip consistency for ConsensusMessage.
+// If arbitrary bytes successfully deserialize into a ConsensusMessage,
+// re-serializing and deserializing again must produce identical bytes.
+// A mismatch indicates a serialization/deserialization asymmetry bug.
+fuzz_target!(|data: &[u8]| {
+    let Ok(msg) = wincode::deserialize::<ConsensusMessage>(data) else {
+        return;
```
akhi3030, line 12: Is it possible that we will be hitting this case most of the times?  Skimming through the documentation, I am wondering if we need to use https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html to get better coverage for the testing in the rest of the function.

fuzz/fuzz_targets/deserialize_consensus_message.rs
```diff
@@ -0,0 +1,8 @@
+#![no_main]
+
+use alpenglow::consensus::ConsensusMessage;
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let _res = wincode::deserialize::<ConsensusMessage>(data);
```
akhi3030, line 7: isn't this covered in the above test?

fuzz/fuzz_targets/merkle_proof_verify.rs
```diff
@@ -0,0 +1,49 @@
...
+use alpenglow::crypto::merkle::PlainMerkleTree;
+use libfuzzer_sys::fuzz_target;
+
+// Fuzz Merkle proof verification with structured inputs.
+// Consumes raw bytes to construct a leaf, root hash, proof hashes, and index,
+// then calls both `check_proof` and `check_proof_last` to ensure they never
+// panic on any combination of inputs.
+fuzz_target!(|data: &[u8]| {
+    // Need at least: 1 byte index + 1 byte leaf_len + 32 bytes root + 32 bytes (one proof hash)
+    if data.len() < 66 {
```
akhi3030, line 13: I wonder if we can use https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html to generate random data that is more useful and then we get better coverage here.

fuzz/fuzz_targets/deserialize_consensus_message.rs
qkniep: currently yes, if we adapt the above to be more structured and produce valid consensus messages, then not anymore

fuzz/fuzz_targets/consensus_message_roundtrip.rs
qkniep: Created #351 to explore this.

test.sh
```diff
@@ -34,6 +34,15 @@ sequential_tests () {
 		three_nodes_crash
 }
 
+fuzz_tests () {
+	echo "🧪 Running fuzz tests!"
+	sleep 1
+	for target in $(cargo +nightly fuzz list); do
```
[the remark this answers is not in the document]
qkniep: done

check.sh
```diff
@@ -5,4 +5,5 @@ cargo build --all-targets --all-features \
 	&& cargo +nightly fmt --check \
 	&& cargo doc --no-deps --document-private-items \
 	&& cargo machete \
+	&& cargo +nightly fuzz build \
 	&& ./test.sh ci
```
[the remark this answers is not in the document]
qkniep: will handle in follow-up

PR #300, opened 2026-01-13
fix: check cert stake threshold

### LINKED ISSUE #299: Check cert threshold
When checking certificate validity we currently only check that the signature is valid, not that the necessary stake threshold is reached.

qkniep:
Previously, when adding a certificate to the pool we only checked that the aggregate signature was valid, but not that the signers together reach the required stake threshold for the certificate type. This is fixed here by adding a `check_threshold()` method to the certificate types, analogous to `check_sig()`.

Closes #299.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "async-lock"
-version = "3.4.1"
+version = "3.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fd03604047cee9b6ce9de9f70c6cd540a0520c813cbd49bae61f33ab80ed1dc"
+checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
 dependencies = [
  "event-listener",
  "event-listener-strategy",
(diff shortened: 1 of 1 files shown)


src/consensus/cert.rs
```diff
@@ -36,6 +36,18 @@ pub enum Cert {
 }
 
 impl Cert {
+    /// Checks that the stake threshold is met.
+    #[must_use]
+    pub fn check_threshold(&self, validators: &[ValidatorInfo]) -> bool {
```
akhi3030: nit: `validators` as an argument here doesn't feel descriptive enough.  I initially thought that it is the list of the validators that signed the cert.  Should we rename it to `epoch_validators` maybe?  Alternatively, the input to the function can also be `&EpochInfo`?

An advantage of using `&EpochInfo` might be that we could store the `total_stake` in there as that would not change during the epoch and then we do not need to recompute it for validating each cert.

src/consensus/cert.rs
```diff
@@ -466,6 +530,18 @@ impl FastFinalCert {
...
+        let total_stake: Stake = validators.iter().map(|v| v.stake).sum();
+        let stake: Stake = validators
+            .iter()
+            .filter(|v| self.agg_sig.is_signer(v.id))
+            .map(|v| v.stake)
+            .sum();
+        stake >= (total_stake * 4).div_ceil(5)
```
akhi3030, line 565: nit: We could consider doing something like https://github.com/anza-xyz/alpenglow/blob/394188b25e46463723f176d8aa0bde9e1c496d48/votor-messages/src/fraction.rs#L7 or using the fraction crate if we wanted to define the thresholds as global consts or something similar.  I am fine with this as we don't expect the thresholds to really ever change but might be slightly more readable.

Alternatively, it can be useful to add a comment like: `// FastFinalCerts need at least 80% stake`.  It can help the readers parse the fractions as percentages and relay back to what they read in the paper.

akhi3030 (approved):
LGTM!  Added some minor nit comments so disabled auto merge.

src/consensus/cert.rs
qkniep: Good point. Decided to pass `EpochInfo`, which already has a `total_stake()` method. So, the optimization of calculating `total_stake` only once per epoch can be made later without affecting callers.

src/consensus/cert.rs
qkniep: I feel like the fractions are overkill for now. Created an issue to consider this again and added comments like you mentioned.

PR #340, opened 2026-04-12
refactor: add integer fraction type, use for quorums

### LINKED ISSUE #302: Maybe introduce fraction type
We could consider adding a fraction type for exact fractional comparisons without the use of floating point numbers, similar to https://github.com/anza-xyz/alpenglow/blob/394188b25e46463723f176d8aa0bde9e1c496d48/votor-messages/src/fraction.rs#L7. This could be especially useful for defining certain fractions, like the stake thresholds, as `const`s, e.g. it would allow writing something like
```rust
const FAST_FINAL_THRESHOLD: Fraction = Fraction::new(4, 5);
```

qkniep:
Closes #302.

### DIFF
--- a/src/consensus/epoch_info.rs
+++ b/src/consensus/epoch_info.rs
@@ -36,7 +36,11 @@ impl EpochInfo {
             );
         }
         let total_stake = validators.iter().map(|v| v.stake).sum();
-        Self { own_id, validators, total_stake }
+        Self {
+            own_id,
+            validators,
+            total_stake,
+        }
     }
 
     /// Gives the validator info for the given validator ID.
--- a/src/consensus/epoch_info.rs
+++ b/src/consensus/epoch_info.rs
@@ -1,6 +1,9 @@
 // Copyright (c) Anza Technology, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::consensus::{
+    QUORUM_THRESHOLD, STRONG_QUORUM_THRESHOLD, WEAK_QUORUM_THRESHOLD, WEAKEST_QUORUM_THRESHOLD,
+};
 use crate::types::SLOTS_PER_WINDOW;
 use crate::{Slot, Stake, ValidatorId, ValidatorInfo};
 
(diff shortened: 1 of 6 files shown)


src/types/fraction.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Clone, Copy, Debug)]
+pub struct Fraction {
+    numerator: u64,
+    denominator: u64,
+}
+
+impl Fraction {
+    /// Creates a new fraction. Panics if `denominator` is zero.
+    #[must_use]
+    pub const fn new(numerator: u64, denominator: u64) -> Self {
```
akhi3030: Should we use `NonZeroU64` for the denominator maybe and then this will not panic.

src/consensus/epoch_info.rs
```diff
@@ -1,20 +1,29 @@
...
 use crate::types::SLOTS_PER_WINDOW;
 use crate::{Slot, Stake, ValidatorId, ValidatorInfo};
 
-/// Epoch-specfic validator information.
+/// Epoch-specific validator information.
 #[derive(Clone, Debug)]
 pub struct EpochInfo {
     pub(crate) own_id: ValidatorId,
     pub(crate) validators: Vec<ValidatorInfo>,
+    total_stake: Stake,
```
akhi3030, line 15: nit: As `validators` field is exposed, it is possible to accidentally manipulate it so that `total_stake` is not correct anymore.  

Maybe we add an accessor for `validator` instead?

I suppose, the same problem also existed already for `own_id`.  As the struct should hold that `own_id` exists in `validators`.  So maybe we can live with this or fix it in future depending on how we are commonly using these fields.

src/consensus/epoch_info.rs
```diff
@@ -1,20 +1,29 @@
...
 }
 
 impl EpochInfo {
     /// Creates a new `EpochInfo` instance with the given validators.
-    pub const fn new(own_id: ValidatorId, validators: Vec<ValidatorInfo>) -> Self {
-        Self { own_id, validators }
+    pub fn new(own_id: ValidatorId, validators: Vec<ValidatorInfo>) -> Self {
+        let total_stake = validators.iter().map(|v| v.stake).sum();
```
akhi3030, line 38: nit: not part of this PR but should we assert that `own_id` exists in `validators`.

qkniep: Started #344 for this.

src/consensus/epoch_info.rs
qkniep: Opened #345 for this.

akhi3030 (approved):

PR #349, opened 2026-04-12
refactor: split generic epoch info and validator epoch info

qkniep:
So far, we only had a type `EpochInfo` that contains the global state of the validator set for a specific epoch, but also always requires an `own_id` to be set. We split this into an `EpochInfo` type that only holds the global state and a `ValidatorEpochInfo` type that wraps the `EpochInfo` and corresponds to a specific validator.

### DIFF
--- a/src/consensus/epoch_info.rs
+++ b/src/consensus/epoch_info.rs
@@ -9,7 +9,7 @@ use crate::{Slot, Stake, ValidatorId, ValidatorInfo};
 /// Shared epoch information, identical across all validators.
 ///
 /// Contains the validator set and derived data for one epoch.
-/// Constructed once per epoch and shared via [`Arc`].
+/// Constructed once per epoch and shared via [`std::sync::Arc`].
 #[derive(Clone, Debug)]
 pub struct EpochInfo {
     pub(crate) validators: Vec<ValidatorInfo>,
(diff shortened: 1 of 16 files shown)


src/consensus/epoch_info.rs
```diff
@@ -104,6 +104,36 @@ impl EpochInfo {
...
+    }
+
+    /// Returns our own validator ID.
+    #[must_use]
+    pub fn own_id(&self) -> ValidatorId {
+        self.own_id
+    }
+}
+
+impl Deref for ValidatorEpochInfo {
```
akhi3030: nit: I would prefer to have an accessor.  When reading code (especially on github without an IDE), then it can get a bit confusing to figure out what what a type is dereferencing to when it isn't the usual suspect.

src/consensus/epoch_info.rs
```diff
@@ -1,33 +1,40 @@
...
+    pub(crate) validators: Vec<ValidatorInfo>,
     total_stake: Stake,
 }
 
+/// Per-validator epoch information, wrapping shared [`EpochInfo`].
+///
+/// Adds the node's own identity on top of the shared epoch data.
+#[derive(Clone, Debug)]
+pub struct ValidatorEpochInfo {
+    pub(crate) own_id: ValidatorId,
```
akhi3030:
```suggestion
    own_id: ValidatorId,
```

We have an accessor and we do not want users to accidentally update this.

akhi3030 (approved):

PR #352, opened 2026-04-12
refactor: cleanup sampling strategies API

### LINKED ISSUE #256: Clean up `SamplingStrategy` APIs
There are some problems with the current API, specifically:
- Strategies that only make sense for quorums, like `FaitAccompli1Sampler`, need to implement `sample()` to sample a single node.
- Right now `sample_multiple()` always takes the number of samples `k` as parameter, although some strategies, like `FaitAccompli1Sampler`, need to be initialized for a specific `k`. Either the samplers should be instantiated for a specific quorum size, and then `sample_quorum()` should always yield a quorum of that size, or these samplers should handle a way of initializing more than one quorum size (or even lazily initializing quorums when `sample_multiple()` is called with a new value for `k`).

qkniep:
This is a change to the sampling strategies API I wanted to clean up for a while. It adds a `QuorumSamplingStrategy` trait for sampling fixed size quorums, whereas the regular `SamplingStrategy` trait tells you how to sample a single node. These two traits are bridged by the `IidQuorumSampler<S>`, which for any `S: SamplingStrategy` implements i.i.d. sampling of a quorum.

This should be cleaner than the previous abstraction where you for example could use FA1 to sample a single node, which was explicitly implemented as a fallback to `StakeWeightedSampler` just to make sense of it.

Closes #256.

### DIFF
--- a/src/bin/node.rs
+++ b/src/bin/node.rs
@@ -12,7 +12,7 @@ use alpenglow::consensus::{Alpenglow, ConsensusMessage, EpochInfo, ValidatorEpoc
 use alpenglow::crypto::aggsig;
 use alpenglow::crypto::signature::SecretKey;
 use alpenglow::disseminator::Rotor;
-use alpenglow::disseminator::rotor::StakeWeightedSampler;
+use alpenglow::disseminator::rotor::{IidQuorumSampler, StakeWeightedSampler};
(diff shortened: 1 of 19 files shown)


akhi3030 (approved):
i am not super familiar with this code.  Approving to allow merging.  Let me know if you do want a proper review.

PR #356, opened 2026-04-13
refactor: direct path for leader to insert pre-validated shreds

### LINKED ISSUE #216: Perf: leader should have a fast path to add a `ValidatedShred` to the blockstore
Here the leader turns a `ValidatedShred` into a `Shred` just to then validate it again within `add_shred_from_disseminator`. Instead, we could probably give the leader a shortcut to `blockstore.disseminated.add_validated_shred()`.

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/pull/186#discussion_r2371719453_

qkniep:
Closes #216.

### DIFF
--- a/src/consensus/block_producer.rs
+++ b/src/consensus/block_producer.rs
@@ -358,7 +358,7 @@ where
                 .blockstore
                 .write()
                 .await
-                .add_validated_shred_from_leader(s)
+                .add_own_shred_as_leader(s)
                 .await;
             if let Ok(Some(block_info)) = block {
                 assert!(maybe_block_hash.is_none());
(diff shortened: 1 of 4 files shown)


akhi3030 (approved):

qkniep:
There seems to be a subtle bug, which our tests fortunately caught. It's probably related to the fact that we do not populate the cached Merkle root field in the blockstore when going through `add_validated_shred()` rather than `add_shred()`.

qkniep:
It was indeed because the Merkle root of the added shreds was not populated in the blockstore. The fix I chose is to move populating the entry from within `ValidatedShred::try_new()` into `slot_block_data::BlockData::add_validated_shred()`, which also makes the interface of `ValidatedShred::try_new()` easier to use (it replaces the `Entry<SliceIndex, SliceRoot>` with a simple `Option<&SliceRoot>`.

akhi3030:
nice catch of the bug... makes me happy about our testing.  I also like the change to the `ValidatedShreds`.

PR #358, opened 2026-04-13
refactor: split slice types (before shredding, after deshredding)

### LINKED ISSUE #80: Different types for a slice before and after shredding
It seems it's kind of the opposite: the Merkle root only appears in `struct Shred` outside of `ShredPayload`. I think it makes sense, because it only can be added after shredding the slice and is thus different from all the other information in the slice (header and payload), which is part of what is initially being shredded. One way we could maybe improve that in the future is to have separate types for a slice before shredding and after de-shredding.

_Originally posted by @qkniep in https://github.com/qkniep/alpenglow/pull/60#discussion_r2263769593_

qkniep:
Closes #80.

### DIFF
--- a/src/types/slice.rs
+++ b/src/types/slice.rs
@@ -18,8 +18,9 @@ use crate::{BlockId, Slot};
 ///
 /// It corresponds to a single batch of data that the leader is about to disseminate.
 /// During shredding, a slice is turned into multiple shreds.
-/// The result of de-shredding is a [`DeshredSlice`], which additionally carries
-/// the Merkle root that is only computable after shredding.
+///
+/// Deshredding results in a [`DeshredSlice`] instead.
+/// It carries the Merkle root, which is only computable after shredding.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Slice {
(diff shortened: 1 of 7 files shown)


src/types/slice.rs
```diff
@@ -100,6 +83,49 @@ impl Slice {
...
 }
 
+/// A slice recovered after deshredding.
+///
+/// Unlike [`Slice`], this type carries the Merkle root over the slice's shreds,
+/// which is only computable after shredding and is verified during deshredding.
+///
+/// All [`Slice`] fields and methods are accessible directly via [`Deref`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct DeshredSlice {
```
akhi3030, line 93: The name sounds a bit like an action to perform i.e. to deshred a slice.  It is a bit mouthful but I wonder if `RecoveredSlice` or `ReconstructedSlice` would be a better name?

akhi3030 (approved):
Other than a suggestion for the name, LGTM.

src/types/slice.rs
qkniep: Good point. Opened #359.

PR #362, opened 2026-04-13
feat: add NACK message to repair

### LINKED ISSUE #145: Add negative response to repair
Again, probably future work.  If we ignore requests that we cannot satisfy, the the requester will have to wait till timeout to try again.  Does it make sense to send a response so that they can retry faster?

_Originally posted by @akhi3030 in https://github.com/qkniep/alpenglow/pull/106#discussion_r2304657642_

qkniep:
Closes #145.

### DIFF
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -179,7 +179,9 @@ where
             }
             RepairRequestType::Shred(block_id, slice, shred_index) => {
                 let blockstore = self.blockstore.read().await;
-                let shred = blockstore.get_shred(block_id, *slice, *shred_index).cloned()?;
+                let shred = blockstore
+                    .get_shred(block_id, *slice, *shred_index)
+                    .cloned()?;
                 Some(RepairResponse::Shred(
                     request.req_type.clone(),
                     shred.into_shred(),
--- a/src/repair.rs
+++ b/src/repair.rs
@@ -144,8 +144,10 @@ where
     /// Otherwise, a [`RepairResponse::Nack`] is sent back to the requester.
     async fn answer_request(&self, request: RepairRequest) -> std::io::Result<()> {
         trace!("answering repair request: {request:?}");
-        let nack = RepairResponse::Nack(request.req_type.clone());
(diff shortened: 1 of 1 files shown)


src/repair.rs
```diff
@@ -134,44 +140,53 @@ where
...
     ///
-    /// If we do not have the necessary information in blockstore, the request is ignored.
-    /// Otherwise, the correct response is sent back to the sender of the request.
+    /// If we have the information in blockstore, it is sent to the requester.
+    /// Otherwise, a [`RepairResponse::Nack`] is sent back to the requester.
     async fn answer_request(&self, request: RepairRequest) -> std::io::Result<()> {
         trace!("answering repair request: {request:?}");
-        let response = match &request.req_type {
+        let nack = RepairResponse::Nack(request.req_type.clone());
+        let response = self.try_build_response(&request).await.unwrap_or(nack);
```
akhi3030:
```suggestion
        let response = self.try_build_response(&request).await.unwrap_or(RepairResponse::Nack(request.req_type));
```

This seems to work and then we do not unnecessarily build the nack if we do not need it.

src/repair.rs
```diff
@@ -293,6 +308,10 @@ where
         };
 
         match response {
+            RepairResponse::Nack(req_type) => {
+                debug!("received NACK for repair request {req_type:?}, retrying immediately");
+                self.send_request(req_type).await.unwrap();
```
akhi3030, line 315: should we check if we had actually made a request for `req_type` before.  Someone could send us a spurious nack to get us to do additional work.

src/repair.rs
qkniep: good point, I think for that we need to use `unwrap_or_else` not `unwrap_or` though

src/repair.rs
qkniep: we already check that above the `match` and early return if we didn't

akhi3030 (approved):

PR #365, opened 2026-04-14
refactor: split `VotorEvent` into per-sender event types

### LINKED ISSUE #39: Consider splitting up `VotorEvent`
It might be reasonable to split up `VotorEvent` into multiple enums, since it already encompasses three entirely separate subsets of events: The events sent from (a) blockstore, (b) pool, and (c) Votor itself.

_Originally posted by @akhi3030 in https://github.com/qkniep/alpenglow/pull/30#discussion_r2216049928_

qkniep:
Closes #39.


src/consensus/pool.rs
```diff
@@ -25,14 +25,51 @@ use tokio::sync::oneshot;
...
+        slot: Slot,
+        parent_slot: Slot,
+        parent_hash: BlockHash,
+    },
+    /// The given block has reached the safe-to-notar status.
+    SafeToNotar(Slot, BlockHash),
+    /// The given slot has reached the safe-to-skip status.
+    SafeToSkip(Slot),
+    /// New certificate created in pool (should then be broadcast by Votor).
+    CertCreated(Box<Cert>),
```
akhi3030, line 52: we already had this in a `Box` earlier so not really part of this PR.

I think we should. consider not putting this in a `Box`.  The size of the enum will be larger but we will do fewer `memove`s when we construct and dereference this.

src/consensus/votor.rs
```diff
@@ -149,99 +137,119 @@ impl<A: All2All> Votor<A> {
...
-                }
-                VotorEvent::CertCreated(cert) => {
-                    match cert.as_ref() {
-                        Cert::Notar(_) => {
-                            self.block_notarized
-                                .insert(cert.slot(), cert.block_hash().cloned().unwrap());
-                            self.try_final(cert.slot(), cert.block_hash().cloned().unwrap())
-                                .await;
+        loop {
+            tokio::select! {
```
akhi3030, line 141: nit: this is a very large block of code now.  It might be worth breaking it out into functions to make readability easier.  

Another reason is that `rustfmt` struggles for code inside macros so it is hard to get formatting right.

src/consensus/votor.rs
```diff
@@ -382,28 +371,34 @@ mod tests {
 
     type A2A = TrivialAll2All<SimulatedNetwork<ConsensusMessage, ConsensusMessage>>;
 
-    async fn start_votor() -> (A2A, mpsc::Sender<VotorEvent>, EpochInfo) {
+    async fn start_votor() -> (
```
akhi3030, line 381: in future PRs, we can consider introducing a `struct TestContext` that stores all the data being returned here.  This makes refactoring tests easier when we add / remove fields.

src/repair.rs
```diff
@@ -527,8 +528,9 @@ mod tests {
         );
         tokio::spawn(async move {
             repair.repair_loop(repair_rx).await;
-            // keep votor_rx alive
-            drop(votor_rx);
+            // keep event receivers alive so sends from blockstore/pool don't fail
+            drop(_blockstore_rx);
+            drop(_pool_rx);
```
akhi3030:
```suggestion
            drop(blockstore_rx);
            drop(pool_rx);
```

I think these don't need the `_` prefix.

src/test_utils.rs
```diff
@@ -133,18 +131,44 @@ pub fn create_random_block(slot: Slot, num_slices: usize) -> Vec<Slice> {
     slices
 }
 
-/// Asserts that two [`VotorEvent`]s are equal.
+/// Asserts that two [`BlockstoreEvent`]s are equal.
 ///
 /// Panics if they are not equal.
-pub fn assert_votor_events_match(ev0: VotorEvent, ev1: VotorEvent) {
+pub fn assert_blockstore_events_match(ev0: BlockstoreEvent, ev1: BlockstoreEvent) {
```
akhi3030: would deriving `PartialEq, Eq` be better instead.?

src/consensus/pool.rs
qkniep: Created #375 for this.

src/consensus/votor.rs
qkniep: good point, this is a terrible block of code

src/repair.rs
qkniep: yes, we don't get clippy warnings if we change this
my thinking on why this might be a good idea was that it highlights that these are not actually used

src/test_utils.rs
qkniep: I thought there was a reason for not doing this, will try it again

src/consensus/votor.rs
qkniep: Created #376 for this.

src/consensus/votor.rs
qkniep: factored these out into `handle_pool_event`, `handle_blockstore_event`, and `handle_timeout_event`, which was now very natural

src/test_utils.rs
qkniep: ok, this just worked, not quite sure why we didn't have this before

akhi3030 (approved):

PR #382, opened 2026-04-17
refactor: use `TestContext` in more tests

### LINKED ISSUE #376: Introduce `struct TestContext`
in future PRs, we can consider introducing a `struct TestContext` that stores all the data being returned here.  This makes refactoring tests easier when we add / remove fields.

_Originally posted by @akhi3030 in https://github.com/qkniep/alpenglow/pull/365#discussion_r3092581914_

qkniep:
Followup to #379. Closes #376.


src/consensus/blockstore.rs
```diff
@@ -459,22 +471,21 @@ mod tests {
 
     #[tokio::test]
     async fn store_one_slice_block() -> Result<()> {
+        let mut ctx = setup();
         let slot = Slot::genesis().next();
-        let (tx, _rx) = mpsc::channel(100);
-        let (sk, mut blockstore) = test_setup(tx);
-        assert!(blockstore.slot_data(slot).is_none());
+        assert!(ctx.blockstore.slot_data(slot).is_none());
```
akhi3030, line 475:
```suggestion
        ctx.blockstore.slot_data(slot).unwrap_none();
```

not part of the PR but this style is better for tests.  With the former, if there is a failure, one has to almost always change the test code to see what went wrong.  This way, we get a bit more info without changing the test code so it's a bit more informative.

akhi3030 (approved):
just a minor nit for a future PR otherwise, LGTM

src/consensus/blockstore.rs
qkniep: `unwrap_none()` does not seem to exist, for `Result<T>` I guess we could do `unwrap_err()`, and in cases like this here we could for example do `assert_eq!(ctx.blockstore.slot_data(slot), None);` (but that requires `Debug` and `PartialEq`), another alternative would be `assert_matches!()`

akhi3030: And when I think about it, it makes sense that `unwrap_none()` does not exists.  

`assert_matches!()` works or `assert!(blockstore.slot_data(slot).is_none(), "{:?}", blockstore.slot_data(slot));`  assuming `Debug` is implemented.  But this feels quite verbose.  So `assert_matches!()` seems the cleanest.

PR #388, opened 2026-04-18
refactor: turn busy wait-for-block loop into async wait

qkniep:
Builds on #381.


src/consensus/pool/finality_tracker.rs
```diff
@@ -211,6 +230,15 @@ impl FinalityTracker {
         event.finalized = Some(finalized.clone());
         self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
 
+        let waiters = std::mem::take(&mut self.finalization_waiters);
+        for (threshold, tx) in waiters {
+            if self.highest_finalized_slot >= threshold {
+                let _ = tx.send(());
```
akhi3030: should we `warn` if this fails?

src/consensus/pool/finality_tracker.rs
```diff
@@ -211,6 +230,15 @@ impl FinalityTracker {
         event.finalized = Some(finalized.clone());
         self.highest_finalized_slot = slot.max(self.highest_finalized_slot);
 
+        let waiters = std::mem::take(&mut self.finalization_waiters);
```
akhi3030: I wonder if we can use `Vec::retain` here?

src/consensus/block_producer.rs
```diff
@@ -456,50 +455,40 @@ async fn wait_for_first_slot(
...
-            Either::Left(parent) => {
-                return SlotReady::Ready(parent);
-            }
-            Either::Right(rx) => rx,
-        }
+    let mut rx = match pool.wait_for_parent_ready(first_slot_in_window) {
+        Either::Left(parent) => return SlotReady::Ready(parent),
+        Either::Right(rx) => rx,
+    };
+    drop(pool);
```
akhi3030, line 470: nit: I am not a fan of `drop` and prefer to rely on scopes.

src/consensus/blockstore.rs
```diff
@@ -129,12 +140,14 @@ impl BlockstoreImpl {
...
             votor_channel,
             epoch_info,
+            block_waiters: BTreeMap::new(),
         }
     }
 
     /// Deletes everything before the given `slot` from the blockstore.
     pub fn prune(&mut self, slot: Slot) {
         self.block_data = self.block_data.split_off(&slot);
+        self.block_waiters.retain(|&s, _| s >= slot);
```
akhi3030, line 150: Are we dropping the sending side here and could this cause a panic when we call receive on the channel?

src/consensus/blockstore.rs
```diff
@@ -408,6 +437,18 @@ impl Blockstore for BlockstoreImpl {
...
+
+    fn wait_for_disseminated_block(
+        &mut self,
+        slot: Slot,
+    ) -> Either<BlockHash, oneshot::Receiver<BlockHash>> {
+        if let Some(hash) = self.disseminated_block_hash(slot) {
+            return Either::Left(hash.clone());
+        }
+        let (tx, rx) = oneshot::channel();
+        self.block_waiters.insert(slot, tx);
```
akhi3030: is it possible that multiple contexts try to wait for the block?  Should we assert that insert returns there wasn't already someone waiting?

src/consensus/pool/finality_tracker.rs
qkniep: I added it as a `debug` log, I think `warn` is too high a level for this, as it will be the default case when we produce a block

src/consensus/pool/finality_tracker.rs
qkniep: `extract_if()` does what we want to do

src/consensus/block_producer.rs
qkniep: I think this is how it would look like with scopes, not sure if that's better:
```rust
    let (fin_rx, mut rx) = {
        let mut pool = pool.write().await;

        // if pool has finalized a later slot, return skip, otherwise get a channel to await on
        let fin_rx = match pool.wait_for_finalization_at_or_after(first_slot_in_window) {
            Either::Left(()) => return SlotReady::Skip,
            Either::Right(rx) => rx,
        };

        // if already have parent ready, return it, otherwise get a channel to await on
        let rx = match pool.wait_for_parent_ready(first_slot_in_window) {
            Either::Left(parent) => return SlotReady::Ready(parent),
            Either::Right(rx) => rx,
        };

        (fin_rx, rx)
    };

    let parent_slot = first_slot_in_window.prev();
    let block_rx = {
        // if blockstore has block for previous slot, return it, otherwise get a channel to await on
        let mut bs = blockstore.write().await;
        match bs.wait_for_disseminated_block(parent_slot) {
            Either::Left(hash) => return SlotReady::ParentReadyNotSeen((parent_slot, hash), rx),
            Either::Right(rx) => rx,
        }
    };

```

src/consensus/blockstore.rs
qkniep: for the current use case this is not needed, but you are right to point out that it would currently silently break, added the assertion and a comment

src/consensus/blockstore.rs
qkniep: good catch, same actually holds for the ParentReady channel, both could be dropped, only the finalization channel will never be dropped and will eventually receive a message, thus we should probably make sure to not panic and instead just stop waiting on the other two channels if the sender is dropped

qkniep: this should work:
```rust
tokio::select! {
    Ok(_) = fin_rx => SlotReady::Skip,
    Ok(block) = &mut rx => SlotReady::Ready(block),
    Ok(hash) = block_rx => SlotReady::ParentReadyNotSeen((parent_slot, hash), rx),
    else => panic!("all channels dropped"),
}
```
Then we still have a potential panic in block production as we pass the ParentReady receiver `rx` to `produce_block()`. Additionally, it would make sense to also pass `final_rx` and cancel block production if it fires.

PR #390, opened 2026-04-18
fix: slice size calculation in `produce_slice_payload()`

qkniep:


src/consensus/block_producer.rs
```diff
@@ -396,43 +400,40 @@ where
...
-    let parent_encoded_len =
-        <Option<BlockId> as wincode::SchemaWrite<DefaultConfig>>::size_of(&parent).unwrap();
-    let mut slice_capacity_left = MAX_DATA_PER_SLICE
-        .checked_sub(parent_encoded_len + 8)
-        .unwrap();
-    let mut txs = Vec::new();
+    // reserve space for: parent info, and
+    // 8 bytes for SlicePayload::data length
+    let parent_encoded_len = wincode::serialized_size(&parent).unwrap() as usize;
+    let buffer_space = MAX_DATA_PER_SLICE - parent_encoded_len - 8;
```
akhi3030, line 409: nit: not part of PR but maybe a comment about the -8 will be useful here.

akhi3030 (approved):

src/consensus/block_producer.rs
qkniep: that is what the "8 bytes" in the comment above refers to, any idea how to make it harder to miss?

akhi3030: `// subtract 8 bytes to account for SlicePayload::data length`

Maybe something like that?

PR #391, opened 2026-04-18
fix: panic in simulated network upon shutdown

qkniep:


akhi3030 (approved):
nit: should we print a log error or something before exiting?

PR #434, opened 2026-05-23
fix: error instead of panic on invalid slice, lower decode size limit

qkniep:

PR #508, opened 2026-05-31
fix: handle non-decodable `SlicePayload`

qkniep:


qkniep:
Closed in favor of #434 and #516.

